PR #23, opened 2025-03-05
build(typescript): use rollup for cjs and esm

joe-p:
Goal is to have one package for ESM and CJS. Largely inspired (mostly copy/pasted) from https://github.com/conda/rattler/pull/1093. 

As someone that doesn't come from the web world a lot of this is a bit opaque to me, so I will keep this in draft until I have someone more versed with this kind of setup to review and help setup proper tests for node and browser


joe-p:
Based on Slack conversation, this seems to be a good incremental step. Eventually we'll want to decide where to draw the lines in terms of what we want to support and what we want the entry points to look like

PR #32, opened 2025-03-17
build(swift): use static libraries for swift

joe-p:
Static linking probably makes the most sense for most use cases of our libraries since sharing won't often be needed. It also make it easier to distribute executables to end users (vs also distributing the dylib)


joe-p:
I realized that this significantly increase build time, which will make iterative development/testing more painful. For now closing this. Perhaps in the future we provide a separate (drop-in replacement) package for static linking?

PR #46, opened 2025-03-29
fix(algo_models): use rpath in dynamic libs

joe-p:


joe-p:
Closed in favor of #47

PR #49, opened 2025-04-06
ci(swift): cache simulator

joe-p:


joe-p:
Couldn't find an obvious way to determine where this is downloaded. Closing now and we can loop back once we start focusing on swift

PR #50, opened 2025-04-16
chore: rename algo_models to algokit_transact

joe-p:
Renames the `algo_models` crate to `algokit_transact`. All references to the old name have been updated. 

Also renames master to main in the workflows


crates/algokit_transact/src/lib.rs
```diff
@@ -15,7 +15,7 @@ const ALGORAND_ADDRESS_LENGTH: usize = 58;
 const ALGORAND_PUBLIC_KEY_BYTE_LENGTH: usize = 32;
 
 #[derive(Debug, Error)]
-pub enum AlgoModelsError {
+pub enum AlgokitTransactError {
```
neilcampbell, line 18:
```suggestion
pub enum AlgoKitTransactError {
```
We typically use `AlgoKit` in other repos.

joe-p: Nice catch. Force pushed [e10c345](https://github.com/algorandfoundation/algokit-core/pull/50/commits/e10c34522c51f4b15cc6884cd7a058b9b8934383) with the change Algokit -> AlgoKit

joe-p: Now [742ac11](https://github.com/algorandfoundation/algokit-core/pull/50/commits/742ac113c1b55b1bb2c5310f4ce433d7245069a7)

neilcampbell (approved):

PR #52, opened 2025-04-16
feat: add support for calculating transaction id

neilcampbell:
Added both `raw_id` and `id` methods.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,6 +33,7 @@ name = "algokit_transact"
 version = "0.1.0"
 dependencies = [
  "base32",
+ "base64",
  "pretty_assertions",
  "rmp-serde",
  "rmpv",
--- a/crates/algokit_transact/Cargo.toml
+++ b/crates/algokit_transact/Cargo.toml
@@ -18,3 +18,4 @@ thiserror = "2.0.7"
 
 [dev-dependencies]
 pretty_assertions = "1.4.1"
+base64 = "0.22.1"
--- a/crates/algokit_transact/src/lib.rs
+++ b/crates/algokit_transact/src/lib.rs
@@ -102,11 +102,10 @@ pub trait TransactionId: AlgorandMsgpack {
     fn raw_id(&self) -> Result<[u8; HASH_BYTES_LENGTH], AlgoKitTransactError> {
         let mut hasher = Sha512_256::new();
         hasher.update(self.encode()?);
-        hasher.finalize().try_into().map_err(|_| {
-            AlgoKitTransactError::InputError(
-                "could not convert transaction hash to 32 bytes".to_string(),
-            )
-        })
+
+        let mut hash = [0u8; HASH_BYTES_LENGTH];
+        hash.copy_from_slice(&hasher.finalize()[..HASH_BYTES_LENGTH]);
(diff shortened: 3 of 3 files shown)


crates/algo_models/src/lib.rs
```diff
@@ -100,6 +98,27 @@ pub trait AlgorandMsgpack: Serialize + for<'de> Deserialize<'de> {
     }
 }
 
+pub trait TransactionId: AlgorandMsgpack {
+    fn raw_id(&self) -> Result<[u8; HASH_BYTES_LENGTH], AlgoModelsError> {
+        let mut hasher = Sha512_256::new();
+        hasher.update(self.encode()?);
+        hasher.finalize().try_into().map_err(|_| {
+            AlgoModelsError::InputError(
```
neilcampbell, line 106: Was tossing up if I should create a new error type. It should never be hit.

crates/algo_models/src/lib.rs
```diff
@@ -457,112 +479,196 @@ impl AlgorandMsgpack for SignedTransaction {
...
+            transaction: unsigned_tx.clone(),
+            signature: [0; 64],
+        };
+
+        assert_eq!(payment.id().unwrap(), expected_tx_id);
+        assert_eq!(unsigned_tx.id().unwrap(), expected_tx_id);
+        assert_eq!(signed_tx.id().unwrap(), expected_tx_id);
+    }
+
+    fn example_pay_transaction() -> PayTransactionFields {
```
neilcampbell, line 638: Curious if there is some conventional way in rust to represent test data?
I was thinking about creating a test data module, however figured i'd push something simple for now.

crates/algo_models/src/lib.rs
joe-p: Since hash length is an invariant I would say this would be an acceptable use of `except` instead of `map_err`.

crates/algo_models/src/lib.rs
joe-p: Good question. Not sure what the idiomatic way to do this is. Also worth noting there is a test data JSON file: https://github.com/algorandfoundation/algokit-core/blob/main/crates/algokit_transact_ffi/test_data.json

This file is generated via this script: https://github.com/algorandfoundation/algokit-core/blob/main/scripts/generate_test_data/index.ts

Maybe it would make sense to use that same data in the rust unit tests? 

Regardless, I think what you have now is fine and we can refine test strategy in a future PR

joe-p (commented):
Aside from minor comments, LGTM. Just need to rebase from main now that #50 is merged

crates/algo_models/src/lib.rs
neilcampbell: Yeah I did a bit of research too, and not much came up.
Potentially we could explore using an Object Mother pattern, which could also expand to include some test data builders.

It probably does make sense to share the test data.

For now I'll leave as is.

neilcampbell:
> Aside from minor comments, LGTM. Just need to rebase from main now that https://github.com/algorandfoundation/algokit-core/pull/50 is merged

@joe-p Done, ready for another look.

joe-p (approved):

PR #54, opened 2025-04-17
feat: added txnid ffi endpoints

lempira:
~~This PR is dependent on #52. Will move out of draft mode, once it's merged.~~

This PR adds the FFI implementation of the the newly added `txn.id` and `txn.raw_id` methods.

### DIFF
--- a/crates/algokit_transact_ffi/src/lib.rs
+++ b/crates/algokit_transact_ffi/src/lib.rs
@@ -1,4 +1,4 @@
-use algokit_transact::AlgorandMsgpack;
+use algokit_transact::{AlgorandMsgpack, TransactionId};
 use ffi_macros::{ffi_func, ffi_record};
 use serde::{Deserialize, Serialize};
 use serde_bytes::ByteBuf;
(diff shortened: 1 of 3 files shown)


joe-p:
#52 has been merged

crates/algokit_transact_ffi/src/lib.rs
```diff
@@ -390,7 +390,7 @@ impl TryFrom<AssetTransferTransactionFields> for algokit_transact::AssetTransfer
 impl TryFrom<algokit_transact::Transaction> for Transaction {
     type Error = AlgoKitTransactError;
 
-    fn try_from(tx: algokit_transact::Transaction) -> Result<Self, Self::Error> {
+    fn try_from(tx: algokit_transact::Transaction) -> Result<Self, AlgoKitTransactError> {
```
joe-p, line 393: Was this giving you a problem or is this change for clarity? I believe the `type Error = AlgoKitTransactError` on the previous line means the behavior is the same either way.

joe-p (approved):
LGTM. One question but nothing needs to be changed

crates/algokit_transact_ffi/src/lib.rs
```diff
@@ -535,4 +551,65 @@ mod tests {
         let tx_type = get_encoded_transaction_type(&encoded).unwrap();
         assert_eq!(tx_type, TransactionType::Payment);
     }
+
+    // Example transaction modeled from algokit_transact::tests
+    fn example_pay_transaction_ffi() -> Transaction {
```
neilcampbell, line 556: Let's chat about an approach to sharing test data.

lempira: Yeah, I was looking at how some other repos do it. I'll post my findings in the chat.

crates/algokit_transact_ffi/src/lib.rs
```diff
@@ -493,9 +493,25 @@ pub fn address_from_string(address: &str) -> Result<Address, AlgoKitTransactErro
         .map_err(|e| AlgoKitTransactError::EncodingError(e.to_string()))
 }
 
+/// Get the raw 32-byte transaction ID for an encoded transaction.
+#[ffi_func]
+pub fn get_transaction_raw_id(encoded_tx: &[u8]) -> Result<Vec<u8>, AlgoKitTransactError> {
+    let tx = algokit_transact::Transaction::decode(encoded_tx)?;
+    let raw_id = tx.raw_id()?;
```
neilcampbell, line 500: Calling `raw_id` re-encodes the transaction to msgpack before generating the hash.

I'm wondering if it's better to only support Transaction, which would be as below:

```rs
#[ffi_func]
pub fn get_transaction_raw_id(tx: Transaction) -> Result<Vec<u8>, AlgoKitTransactError> {
    let ctx: algokit_transact::Transaction = tx.try_into()?;
    let raw_id = ctx.raw_id()?;
    Ok(raw_id.to_vec())
}
```

PR #56, opened 2025-04-22
chore: adding ts and py test for transaction id

neilcampbell:
I'm using the test transaction data that is included in the json file, however I think it makes sense to share the test data we'll expose in the object mother / test data factory.


lempira (commented):
Looks good, but I made a change in [another PR](https://github.com/algorandfoundation/algokit-core/pull/54#discussion_r2053404338) based on your previous comment that will affect this PR.

PatrickDinh (approved):

PR #57, opened 2025-04-22
refactor: changed the algokit_transact crate dir structure

lempira:
AK-243

In order to break up `'lib.rs`, I updated the algokit_transact crate dir structure to following structure

src/lib.rs # Main entry point, re-exports modules
src/error.rs # AlgoKitTransactError definition 
src/traits.rs # AlgorandMsgpack and TransactionId traits
src/utils.rs # Helper functions (sort_msgpack_value, is_zero*, is_empty*, etc.)
src/address.rs # Address implementation
src/constants.rs # Constants used in the crate
src/transactions/mod.rs # Re-exports transaction modules
src/transactions/common.rs # TransactionHeader and shared structures
src/transactions/payment.rs # Payment transaction
src/transactions/asset_transfer.rs # Asset transfer transaction

I originally updated to not break up the transaction to individual files, but I thought that breaking it up made it cleaner. You can clearly see each transaction and everything related to that txn in each file. Could be convinced otherwise, tho.


crates/algokit_transact/src/transactions/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+                let mut stxn: SignedTransaction = rmp_serde::from_slice(bytes)?;
+
+                stxn.transaction = txn;
+
+                return Ok(stxn);
+            }
+            _ => {
+                return Err(AlgoKitTransactError::InputError(format!(
+                    "expected signed transaction to be a map, but got a: {:#?}",
+                    value.type_id()
```
lempira, line 91: I had type errors with `value.type_id()` until I brought `use std::any::Any;` into scope. I don't know why this is worked, tho. I couldn't tell what the `Any` trait had that fixed the error. 
Any ideas, anyone?

neilcampbell (approved):
LGTM

PR #62, opened 2025-04-28
chore: change transact crate structs to align with ffi crate structs

neilcampbell:
Adjusted the transact structs to more closely align to the ffi transaction crate structs.
This should result in a more consistent feel when using the transact rust crate vs ffi generated package.

Hopefully this should also help simplify the test data sharing approach.


neilcampbell:
Yeah this one definitely has pros and cons. For the most part i'm pro the idiomatic rust stance.

I guess it's mostly a question of how much we value consistency with the type structures between Rust and the generated languages. In theory this consistency should make documentation, test data and the FFI layer simpler.

I also like that this change has removed the header field, which felt a little awkward to me in the TS/Py types.

Worth also mentioning that I'm planning on adding a `TransactionBuilder` (generated using `derive_builder`) which helps with the ergonomics. It's a common pattern in rust to deal with not having to explicitly supply None to Option fields like below:

```rust
let transactionResult = TransactionBuilder::new(
	  String::from("testnet-v1.0"),
	  [0; 32]
  )
  .payment(PaymentTransactionFieldsBuilder::default()
      .amount(101000)
      .receiver(bob)
  )
  .sender(jane)
  .fee(1000)
  .first_valid(50659540)
  .last_valid(50660540)
  .build()
```

> memory-efficiency of tagged enums

Curious about this one, as I was under the impression that the memory footprint of a tagged union in rust was sized to the largest variant?

joe-p:
> Curious about this one, as I was under the impression that the memory footprint of a tagged union in rust was sized to the largest variant?

Correct, but an `Option<T>` is the size of `T` + 1 byte meaning the size of a struct with multiple `Option`s is the *sum* of all the types. 

I think the more important part though is the compiler-checks we can get with `enum` (i.e. ensuring all branches exist in a `match`) and pattern matching.

neilcampbell:
> Correct, but an Option<T> is the size of T + 1 byte meaning the size of a struct with multiple Options is the sum of all the types.

Ah yes, I see.

joe-p:
Closing this based on the team's conversation today regarding idiomatic rust vs FFI alignment. At this time we think it's worthwhile to keep Rust implementations idiomatic, even at the expense of some documentation pains (i.e. harder to integrate into dev portal)

PR #63, opened 2025-04-28
feat: initial oas generators support; algod http clients with tests covering transaction endpoints; refinements in python ffi package

aorumbayev:
To be expanded


joe-p:
Haven't reviewed in-depth yet, but would it be possible make the Python changes in a separate PR? We have merge conflicts anyways after #61 and having that work in a separate PR will make it a bit easier to review and iterate.

aorumbayev:
@joe-p moved python ffi changes to https://github.com/algorandfoundation/algokit-core/pull/65

api/README.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+The converted spec will be available at `specs/algod.oas3.json`.
+
+### Generate API Clients
+
+Generates TypeScript and Python API clients based on the OpenAPI spec:
+
+```bash
+bun run generate-clients
```
PatrickDinh, line 34:
```suggestion
bun run generate-algod-clients
```

api/scripts/generate-algod-clients.ts
```diff
@@ -0,0 +1,110 @@
...
+import { execSync } from "child_process";
+import { join } from "path";
+import { mkdirSync, existsSync, copyFileSync } from "fs";
+
+// Parse command line arguments
+const args = process.argv.slice(2);
+const targetIndex = args.indexOf('--target');
+const target = targetIndex !== -1 ? args[targetIndex + 1]?.toLowerCase() : 'all';
+
+// TODO: use full spec, for now just txns for testing purposes
```
PatrickDinh, line 12: I think this TODO can be removed now

api/scripts/generate-algod-clients.ts
```diff
@@ -0,0 +1,110 @@
...
+        console.log(
+            "ensuring ignore rules are propagated before client generation"
+        );
+        const destPath = join(outputDir, ".openapi-generator-ignore");
+        copyFileSync(ignoreFilePath, destPath);
+    }
+}
+
+function generateTypescriptClient() {
+    console.log("Generating TypeScript client...");
```
PatrickDinh, line 50: Should we delete the `TYPESCRIPT_OUTPUT` before generating?
Same comment for python

PatrickDinh (approved):

joe-p:
Could we add extension to the moustache template files to make them a bit easier to digest? For example, instead of `conftest.moustache` it could be `conftest.py.moustache`

api/oas_templates/typescript/package.mustache
```diff
@@ -0,0 +1,66 @@
...
+      "import": "./dist/esm/*.js",
+      "require": "./dist/cjs/*.js"
+    }
+  },
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "test": "vitest",
+    "build": "npm run clean && npm run build:esm && npm run build:cjs",
```
PatrickDinh, line 48: hey @aorumbayev, sorry I missed this in my first round of review. Do you think that we should change this to bun instead of npm to align with the rest of the code base?

aorumbayev:
@joe-p adding extensions to mustache templates, while possible with mustache templating in general, wouldn't fit well to our use case. The base template was generated via oas generator command that stores the base files to the filesystem so developers can then extend them. The problem is that extension mapping for template files included in the base template are all postfixed with .mustache, if we are to preserve the extensions from original base templates we would have to manually update the files and then original generator would threat them as supporting external files instead of recognizing as files from base templates. Whenever potential updates in base templates arrive it would be hard for us to manually keep track of changes and remapping against latest templates we maintain with the file extensions. 

Once we have CD for clients, we can add the generated clients back to version control for tracking output stability, at that point it would be fairly easy to traceback from generated file back to mustache template by the filename. In most cases those would be custom template files for tests and those also contain explicit filename specification in the spec files per template folder. https://github.com/algorandfoundation/algokit-core/blob/spike/oas_clients/api/oas_templates/python/openapi-config.yaml

.github/workflows/api_clients_ci.yml
```diff
@@ -0,0 +1,128 @@
...
+          retention-days: 1
+
+      - name: Generate Python client
+        working-directory: api
+        run: bun run scripts/generate-algod-clients.ts --target python
+      
+      - name: Upload Python client
+        uses: actions/upload-artifact@v4
+        with:
+          name: python-client
```
neilcampbell, line 47: Is the plan still to publish a seperate client for algod, kmd, indexer?
If so this should probably we named something like `algokit-algod-api-py` and typescript should follow the convention.

neilcampbell (approved):

PR #65, opened 2025-04-29
refactor: python ffi package improvements

aorumbayev:
## Proposed changes

Includes improvements to the Python FFI package (`packages/python/algokit_transact`):

1.  **mypy support and explicit control over content exposed from ffi wheel:** Enhanced the package with proper mypy support by injecting a py.typed marker file. Restructured the module layout to eliminate the redundant nested structure (algokit_transact.algokit_transact). This was achieved by configuring Maturin in pyproject.toml to build the FFI-generated native code as _algokit_transact (module-name = "_algokit_transact"), while adding explicit __init__.py files that selectively re-export symbols. This approach creates a clear separation between autogenerated FFI artifacts and manually curated package components, giving full control over the public API exposed in the final wheel.
2.  **Enhanced Poetry Integration:**
    -   Added a `poetry.toml` file configuring `virtualenvs.in-project = true`. This simplifies the local development workflow by creating the virtual environment within the project directory.
    -   Updated dependencies in `pyproject.toml` and `poetry.lock`, including specifying versions for `cffi` that are compatible with python 3.13 and refining dev dependencies.
3.  **Docs** adding dedicated readme highlighting quick starts on the ffi package along with options on how to work with it within ides using maturin develop command that simplifies developing new tests for the package.


packages/python/algokit_transact/pyproject.toml
```diff
@@ -1,26 +1,28 @@
...
+requires = ["maturin>=1.5,<2"]
 build-backend = "maturin"
 
 [project]
 name = "algokit_transact"
 version = "0.1.0"
-requires-python = ">=3.7"
-dependencies = ["pynacl>=1.5.0"]
+requires-python = ">=3.8,<4.0"
+dependencies = ["pynacl>=1.5.0,<2.0.0", "cffi>=1.17.1"]
```
joe-p, line 9: It just occurred to me that `pynacl` is probably not an actual dependency... I believe it's just used for testing

packages/python/algokit_transact/pyproject.toml
```diff
@@ -1,26 +1,28 @@
...
 
 [tool.maturin]
 manifest-path = "../../../crates/algokit_transact_ffi/Cargo.toml"
+module-name = "_algokit_transact"
```
joe-p, line 13: How come there's a leading underscore here?

joe-p (commented):
Overall LGTM. Will want to wait to merge until CI runs, but right now it's all being skipped due a bug that was fixed in #64. Once #64 is merged we can rerun CI to make sure everything is working

packages/python/algokit_transact/pyproject.toml
aorumbayev: @joe-p as part of injecting the py.typed to add mypy support, i found a few best practices on maturin doc that would allow us to have clear separation between autogenerated ffi bits vs extra python specific code we need to inject https://www.maturin.rs/project_layout.html#import-rust-as-a-submodule-of-your-project which is just py.typed at the moment. Because we have module name matching the name of the python project a resulting wheel with injected py.typed would have resulted in nested structure algokit_transact.algokit_transact. With the underscore we explicitly name and mark the auto generated bits as private module and then re export everything from it from the root init py. 

I did revisit the doc and just pushed an extra tweak to place the _algokit_transact explicitly under algokit_transact module as its highlighted as recommended pattern in their doc. 

In future work this gives us the ability to inject any extra python code and have a clean way to separate between whats python vs autogenerated ffi code.

packages/python/algokit_transact/pyproject.toml
aorumbayev: Removed

joe-p:
CI is failing with `ModuleNotFoundError: No module named 'algokit_transact._algokit_transact'` on both Mac ARM and Linux x86_64 which is strange because the wheel is seemingly being installed.

joe-p:
I wonder if the failure with musl is happening because we removed PyNacl? Maybe related to `pycparser`?

aorumbayev:
@joe-p fixed the issue, had to revert the structure i had initially. Added a detailed comment in pyproject py explaining why we don't use a convention from maturin docs.

joe-p (approved):

PR #66, opened 2025-04-30
Some initial documentation

CiottiGiorgio:
This is some attempt at setting a documentation standard that we like.

We could decide to be more or less verbose, for example.

One thing I'm looking for feedback on is if we want to use the `# Errors` markdown.

For now, most of the core crate is done and this PR aims at doing at least the FFI crate too.


crates/algokit_transact/src/transactions/asset_transfer.rs
```diff
@@ -1,37 +1,70 @@
+//! Asset transfer transaction module for Algokit Core.
```
neilcampbell, line 1:
```suggestion
//! Asset transfer transaction module for AlgoKit Core.
```

crates/algokit_transact/src/transactions/asset_transfer.rs
```diff
@@ -1,37 +1,70 @@
...
     #[serde(skip_serializing_if = "is_zero_addr")]
     #[serde(default)]
     pub receiver: Address,
 
+    /// Optional address of the account that actually holds the asset being transferred.
+    ///
+    /// If provided, this indicates that the transaction is a clawback operation,
+    /// where the sender is the asset clawback address and is forcibly moving assets
+    /// from this account to the receiver.
+    // FIXME: This description is untouched from what Claude wrote.
```
neilcampbell, line 55: Anything additional you want to tweak?

crates/algokit_transact/src/transactions/asset_transfer.rs
```diff
@@ -1,37 +1,70 @@
...
     #[serde(skip_serializing_if = "is_zero_addr_opt")]
     #[serde(default)]
     pub asset_sender: Option<Address>,
 
+    /// Optional address to send all remaining asset units to after the transfer.
+    ///
+    /// If specified, this indicates that the sender is closing out their position in the asset,
+    /// and all remaining units of this asset owned by the sender will be transferred to this address.
+    /// This effectively removes the asset from the sender's account.
+    // FIXME: Same as above.
```
neilcampbell, line 67: Anything additional you want to tweak?

crates/algokit_transact/src/transactions/asset_transfer.rs
```diff
@@ -1,37 +1,70 @@
...
     #[serde(default)]
     pub receiver: Address,
 
+    /// Optional address of the account that actually holds the asset being transferred.
+    ///
+    /// If provided, this indicates that the transaction is a clawback operation,
+    /// where the sender is the asset clawback address and is forcibly moving assets
+    /// from this account to the receiver.
+    // FIXME: This description is untouched from what Claude wrote.
+    // We should assess if asset_sender should be part of a basic asset transfer (probably not).
```
neilcampbell, line 56: I feel like we want to represent all the various raw transaction fields at the core level (like it currently does) and let higher level abstractions like utils take care of splitting based on use case.

crates/algokit_transact/src/transactions/mod.rs
```diff
@@ -46,20 +70,37 @@ impl AlgorandMsgpack for Transaction {
...
 }
 
 impl AlgorandMsgpack for SignedTransaction {
+    /// The prefix used for MessagePack encoding, empty for signed transactions.
     const PREFIX: &'static [u8] = b"";
 
+    /// Decodes MessagePack bytes into a SignedTransaction.
+    ///
+    /// This implementation handles the complexity of determining the transaction type
+    /// correctly during deserialization.
```
neilcampbell, line 94: Is this needed? What's the complexity?

crates/algokit_transact/src/transactions/payment.rs
```diff
@@ -1,27 +1,43 @@
+//! Payment transaction module for Algokit Core.
```
neilcampbell, line 1:
```suggestion
//! Payment transaction module for AlgoKit Core.
```

crates/algokit_transact/src/transactions/asset_transfer.rs
CiottiGiorgio: I would definitely want both be represented. Not sure if utils is the best place for this higher abstraction because I would want it to be where most users are consuming the crates and not make them so optional.
What I mean is that I see this higher abstraction being the default, while the actual representation being something you reach for only in rare cases.
Granted, both should exist.

crates/algokit_transact/src/transactions/asset_transfer.rs
CiottiGiorgio: I _think_ I'm mostly fine with the content. I was just considering if this is the right place to put it but I guess it is.

CiottiGiorgio: I'll replace this FIXME with a reminder to talk about / decide on the transaction higher abstraction.

crates/algokit_transact/src/transactions/mod.rs
CiottiGiorgio: Yeah this was superfluous. Removed.

crates/algokit_transact/src/address.rs
```diff
@@ -55,10 +89,21 @@ impl Address {
...
+    /// A 4-byte array containing the checksum.
     pub fn checksum(&self) -> [u8; ALGORAND_CHECKSUM_BYTE_LENGTH] {
         pub_key_to_checksum(&self.pub_key)
     }
 
+    /// Converts the Address to its standard base32-encoded string representation.
+    ///
+    /// # Returns
+    /// A string containing the base32-encoded address.
+    // FIXME: Address.address? Is this the right name? Is there a canonical way to convert an
```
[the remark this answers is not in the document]
CiottiGiorgio: Yeah that sounds good. I'd also implement `FromStr` since we are at it and see if it works well?

It should be a sensible replacement for `.address()` and `.from_string()` wherever we are using it currently.
I'm also wondering if it handles an implicit casting to `str` as well? I'm thinking of cases where the argument to a query/function is the human readable representation.

crates/algokit_transact/src/transactions/asset_transfer.rs
neilcampbell: Thoughts @lempira and @joe-p.

crates/algokit_transact/src/address.rs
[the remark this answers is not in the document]
CiottiGiorgio: Pointing out PR #75.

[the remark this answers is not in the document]
joe-p: Yeah implementing `Display` probably makes more sense than having `.address()`

neilcampbell (approved):

PR #70, opened 2025-05-02
feat: Add transaction builders and test object mother abstractions

neilcampbell:
- Added builders for creating payment and asset transfer transactions
- Added transaction, transaction header, address, and test data object mothers
- Test data json is generated when running `cargo build` on `algokit_transaction_ffi`
- Renamed `PayTransactionFields` to `PayTransactionFields`
- Renamed `pay_fields` to `payment` in FFI `Transaction` struct
- Renamed `asset_transfer_fields` to `asset_transfer` in FFI `Transaction` struct


.github/workflows/python_uniffi_ci_cd.yml
```diff
@@ -32,7 +32,7 @@ env:
 jobs:
   get_version:
     runs-on: ubuntu-latest
-    outputs:
```
joe-p, line 35: Noticed some formatting was done by your IDE. Not a big deal, but we should probably align on formatting as a team to avoid potentially going back & forth in PRs

.github/workflows/python_uniffi_ci_cd.yml
```diff
@@ -210,7 +210,7 @@ jobs:
         env:
           RUN: |
             cd /workspace/packages/python/${{ inputs.crate }}
-            python${{ env.PYTHON_VERSION }} -m maturin build --release --target ${{ env.TARGET }} --compatibility ${{ env.MANYLINUX }} --features ffi_uniffi
+            python${{ env.PYTHON_VERSION }} -m maturin build --release --target ${{ env.TARGET }} --compatibility ${{ env.MANYLINUX }} --no-default-features --features ffi_uniffi
```
joe-p, line 213: Just to make sure I understand, the `--no-default-features` flag is required here and many other places because the testing feature? Can we simply not make the test-related features default?

crates/algokit_transact/src/test_utils/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+};
+use base64::{prelude::BASE64_STANDARD, Engine};
+use convert_case::{Case, Casing};
+use ed25519_dalek::{Signer, SigningKey};
+use serde::Serialize;
+use serde_json::to_writer_pretty;
+
+pub struct TransactionHeaderMother {}
+impl TransactionHeaderMother {
+    pub fn testnet() -> TransactionHeaderBuilder {
```
joe-p, line 16: I think `testnet` and `mainnet` would be useful methods to have on the builder itself

crates/algokit_transact_ffi/src/lib.rs
```diff
@@ -454,16 +456,16 @@ pub fn get_encoded_transaction_type(bytes: &[u8]) -> Result<TransactionType, Alg
     }
 }
 
-#[ffi_func]
```
joe-p, line 457: Similar comment as I made on YAML RE formatting. We should define what tools we are using

scripts/build/languages/typescript.ts
```diff
@@ -4,7 +4,7 @@ import * as fs from "fs";
 async function wasmPack(
   crate: string,
   target: "web" | "nodejs" | "bundler",
-  cwd: string,
+  cwd: string
```
joe-p, line 7: Also TypeScript formatting

joe-p (approved):
LGTM. Will need to resolve merge conflicts with CI

One common thing I noticed is that there were some diffs caused by formatting for Rust, YAML, and TS so we should probably align on what formatting tools we want to use and ensure they are setup properly in the repo. 

Also there's an opportunity for us to add some abstraction in the builder (i.e. `mainnet` and `testnet` functions) but I don't think we need to focus on that now.

crates/algokit_transact/src/test_utils/mod.rs
neilcampbell: Yeah I was tossing up between here vs the builder.
The 2 reason I chose the object mother (at least for now):
- It's a behaviour that wouldn't be available in the FFI layer, so is more consistent and we wouldn't be using this in the utils libraries.
- If testnet ever needed to be reset (hopefully never), it would require a core lib update.

Ultimately it felt like this capability was an abstraction level up from the core.

.github/workflows/python_uniffi_ci_cd.yml
neilcampbell: Yeah 100% agree.
I'll take care of this in another PR, as there will be be a bunch of whitespace related changes.

.github/workflows/python_uniffi_ci_cd.yml
neilcampbell: Good question. Yeah specifically the test data generation capability, which leverages `algokit_transact/test_utils`.
In answering this I realised I made the assumption that the test_utils features would be statically linked into the built transact ffi package binaries, however upon testing it doesn't appear to be true 😄

Enabling the feature by default actually reduces the wasm binary output size, which is interesting.

I'll update based on your suggestion.

PR #72, opened 2025-05-04
feat: Remove transaction type from the transaction header struct

neilcampbell:
When adding the builders and object mothers I noticed that when creating a transaction it felt a little redundant assigning a specific transaction type inside a transaction header, given that it's nested within the `*TransactionFields` struct and then also wrapped in an enum variant which both specify the transaction type. Additionally it was possible to create a transaction with a mismatched transaction type.

This PR removes the TransactionType enum and leverages the serde enum tag feature. The FFI trait definitions have not changed.

### DIFF
--- a/crates/algokit_transact/src/test_utils/mod.rs
+++ b/crates/algokit_transact/src/test_utils/mod.rs
@@ -3,7 +3,7 @@ use std::fs::File;
 use crate::{
     transactions::{AssetTransferTransactionBuilder, PaymentTransactionBuilder},
     Address, AlgorandMsgpack, Byte32, SignedTransaction, Transaction, TransactionHeaderBuilder,
-    TransactionId, TransactionType,
+    TransactionId,
 };
 use base64::{prelude::BASE64_STANDARD, Engine};
 use convert_case::{Case, Casing};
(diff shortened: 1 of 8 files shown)


crates/algokit_transact/src/test_utils/mod.rs
```diff
@@ -36,42 +37,24 @@ impl TransactionHeaderMother {
...
                     .unwrap(),
             )
+            .fee(1000)
             .to_owned()
     }
 
-    pub fn simple_testnet_payment() -> TransactionHeaderBuilder {
+    pub fn simple_testnet() -> TransactionHeaderBuilder {
         Self::testnet()
-            .transaction_type(TransactionType::Payment)
```
joe-p, line 44: I'm a bit confused on why/how we're removing the transaction type here. Is it payment by default?

joe-p: Ahhh ok that makes sense. I was missing the fact that this was just for the header

neilcampbell:
@joe-p This PR has been rebase to the latest main and should be good to review again.

joe-p (approved):

PR #76, opened 2025-05-11
feat: add ci/cd workflow for the algod api clients

neilcampbell:
Also fixes some issues with the existing CI/CD pipeline jobs.

Depends on #74.


.github/workflows/ci_cd.yml
```diff
@@ -15,12 +15,15 @@ jobs:
...
+      # The API client packages
+      api_packages: ${{ steps.set_api_packages.outputs.api_packages }}
     steps:
       - id: set_ffi_packages
         run: echo 'ffi_packages=["algokit_transact"]' >> $GITHUB_OUTPUT
+      - id: set_api_packages
+        run: echo 'api_packages=["algod_api"]' >> $GITHUB_OUTPUT
 
   typescript_wasm_ci_cd:
-    concurrency: ${{ github.event_name == 'push' && format('cd-{0}', github.ref) || format('ci-typescript_wasm-{0}', github.run_id) }}
```
neilcampbell, line 23: I'm assuming this was added as a mechanism for running these steps sequentially only when running release builds?

Unfortunately concurrency in GH actions doesn't quite work like that and instead returns `Canceling since a higher priority waiting request for '{build}' exists`.

.github/workflows/python_uniffi_ci_cd.yml
```diff
@@ -54,17 +55,24 @@ jobs:
...
+      - name: Convert to Python Version
+        if: steps.semantic_release.outputs.version != ''
+        id: convert_version
+        uses: ./.github/actions/as-python-version
+        with:
+          version: ${{ steps.semantic_release.outputs.version }}
       - name: Update version in pyproject.toml
         if: steps.semantic_release.outputs.version != ''
-        run: sed -i 's/version = .*/version = "${{ steps.semantic_release.outputs.version }}"/' pyproject.toml
+        run: sed -i '0,/version = .*/s//version = "${{ steps.convert_version.outputs.version }}"/' pyproject.toml
```
neilcampbell, line 75: Only replace the first match, as dependencies can (and do) also have a "version" key specified.

aorumbayev (approved):
Worth verifying e2e from a fork otherwise looks good to me

PR #81, opened 2025-05-14
feat: estimate transaction size in the FFI crate

CiottiGiorgio:
I thought this code is common enough we might want it in core.

I am still missing a way to do it in the rust crate. After I do that, the FFI crate should only re-expose what's in the core crate.

Also, I haven't taken the approach of actually signing it as it is a slow operation. I decided to attach an invalid signature, encode, and get the size.

I've also seen a bunch of times just adding `75` to the encoding length without a signature. I am assuming `75` comes up from the field name, the encoding and the signature length.
What approach do we think is best?

### DIFF
--- a/crates/algokit_transact/src/transactions/mod.rs
+++ b/crates/algokit_transact/src/transactions/mod.rs
@@ -61,7 +61,7 @@ impl TransactionId for Transaction {}
 
 impl EstimateTransactionSize for Transaction {
     fn estimate_size(&self) -> Result<usize, AlgoKitTransactError> {
-        return Ok(self.encode()?.len() + ALGORAND_SIGNATURE_ENCODING_INCR);
+        return Ok(self.encode_raw()?.len() + ALGORAND_SIGNATURE_ENCODING_INCR);
     }
 }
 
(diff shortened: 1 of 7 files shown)


neilcampbell:
> I've also seen a bunch of times just adding 75 to the encoding length without a signature. I am assuming 75 comes up from the field name, the encoding and the signature length.

Yeah that's what the JS SDK does to. I think we keep the logic the same.

CiottiGiorgio:
Huh. Interesting. I'll make the change.

crates/algokit_transact/src/transactions/mod.rs
```diff
@@ -59,6 +59,19 @@ impl AssetTransferTransactionBuilder {
 impl AlgorandMsgpack for Transaction {}
 impl TransactionId for Transaction {}
 
+impl EstimateTransactionSize for Transaction {
+    fn estimate_size(&self) -> Result<usize, AlgoKitTransactError> {
```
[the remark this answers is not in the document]
CiottiGiorgio: Fixed

crates/algokit_transact/src/transactions/mod.rs
```diff
@@ -59,6 +59,12 @@ impl AssetTransferTransactionBuilder {
...
```
neilcampbell, line 63: Worth adding a simple test?

neilcampbell (approved):

crates/algokit_transact/src/transactions/mod.rs
CiottiGiorgio: Absolutely

joe-p:
Yeah I imagine 75 is the size of the signature struct and signature itself? This is probably something we can improve by adding parameters to the estimate_size function, but I don't think that necessary right now

joe-p (approved):

PR #93, opened 2025-05-19
Fix bundling for commonjs

PatrickDinh:
Fixes #
- Bundling for commonjs
- Remove the need for top level await

### DIFF
--- a/packages/typescript/algokit_transact/src/esm.ts
+++ b/packages/typescript/algokit_transact/src/esm.ts
@@ -6,4 +6,4 @@ import mod from "../pkg/algokit_transact_ffi_bg.wasm";
 import { initSync } from "../pkg/algokit_transact_ffi";
 
 //@ts-ignore
-await initSync({ module: await mod() });
+initSync({ module: mod() });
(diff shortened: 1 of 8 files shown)


packages/typescript/algokit_transact/src/esm.ts
```diff
@@ -6,4 +6,4 @@ import mod from "../pkg/algokit_transact_ffi_bg.wasm";
 import { initSync } from "../pkg/algokit_transact_ffi";
 
 //@ts-ignore
-await initSync({ module: await mod() });
+initSync({ module: mod() });
```
PatrickDinh, line 9: the `initSync` method is a sync method, we don't need to await it. This means we don't need to support top level await

scripts/build/languages/typescript.ts
```diff
@@ -17,7 +17,7 @@ async function build(crate: string, mode: "esm" | "cjs" | "wasm2js", cwd: string
       await wasmPack(crate, "web", cwd);
       break;
     case "cjs":
-      await wasmPack(crate, "nodejs", cwd);
+      await wasmPack(crate, "web", cwd);
```
PatrickDinh, line 20: For commonjs, we still run wasm-pack targeting the web. If we target node, the output will try to load the wasm file from disk. We don't need this behaviour because the plugin `@rollup/plugin-wasm` will inline the wasm file as base64 string for us.

neilcampbell (approved):

aorumbayev (approved):

PR #95, opened 2025-05-21
feat: initial msgpack encoding/decoding support for Simulate endpoint

aorumbayev:
## Proposed changes

Adds initial full e2e flow supporting msgpack encoding decoding in simulate requests. 

- Merges manual model definitions in rust for simulate request and response objects by combining previous state of this pr (see commit history) and taking into account latest architecture discussions into algokit_transact and algokit_transact_ffi crates
- Expands python and typescript templates to support msgpack encoding and decoding of:
- - SimulateRequest and SimulateResponse models (encoding request, decoding response)
- Adds extra integration tests to validate simulate endpoint using msgpack encoding/decoding via transact crate
- Adds extra integration tests to validate account information endpoints (json only)

## Follow up actions needed

- In separate PR commit to renaming algokit_transact_ffi crate into algokit_models_ffi as per latest architecture discussion 
- In separate PR commit to exposing SignedTransaction object via ffi. Afterwards, test and simplify manual decoding encoding logic for simulate models manually written in this pr and merge. 
- In separate PR, upon merging above two final step is to optimize auto generation of **SELECT** message pack models that we want to support. IMHO we should additionally add support for block endpoint and only add new msgpack models if we see the need or upon requests - evidence from algosdk libraries suggests that primary areas of concern where msgpack is defaulted to is mainly blocks, simulate and pending transaction endpoints.


api/specs/compatibility.md
```diff
@@ -55,3 +55,25 @@
 | GET /v2/experimental                                     | ❌               | N/A           | ✅            | N/A        |
 | GET /v2/devmode/blocks/offset                            | ❌               | N/A           | ✅            | N/A        |
 | POST /v2/devmode/blocks/offset/{offset}                  | ❌               | N/A           | ✅            | N/A        |
+
+Similar to above but focused on abstractions:
+
+| Abstraction                         | Related Endpoints                                                             | Supports msgpack Encoding | Supports msgpack Decoding |
+| ----------------------------------- | ----------------------------------------------------------------------------- | ------------------------- | ------------------------- |
+| Account                             | GET /v2/accounts/{address}                                                    | No                        | Yes                       |
```
aorumbayev, line 63: @neilcampbell @joe-p @PatrickDinh 
I am of an opinion that we should not allow performing decoding for models that only support/are relevant for encoding and vice versa. For instance Account model is mainly returned in response when msgpack is specified explicitly, its relevant only for decoding. Unless you can think of the scenarios where something like that would useful, let me know if you agree that it makes sense to comply with the spec's behaviour and simply throw something like NotImplemented whenever user tries to encode/decode a model that is only meant for decoding/encoding or does not support msgpack at all. 

This table summarizes all abstractions from algod spec that are relevant to msgpack formats along with their associated endpoints and whether they do support  encoding and/or decoding.

aorumbayev, line 63: Fun fact, when you use the accounts endpoint and try to obtain the account info in msgpack format, the returned msgpack object when decoded differs from what algod states in the schema. It seems like schema only prefers the json formats without explicitly documenting the discrepancies on what happens with the same schema structure when in msgpack format. I am not sure why in msgpack format algod does not return the address field, which according to the spec can't be a null object. I tweaked the decoding atm to set an empty string when decoding that msgpack. 

We would need to document this difference somewhere in our own version of the algod spec so its reflected in docstrings. @joe-p are you aware of any other algod related models where schema states in the spec differs from what is returned on the same endpoints for the same schema but under msgpack format?

neilcampbell: That makes sense to me if the OAS doesn't define it as a supported format.

neilcampbell: Rather than setting to an empty string, could we use the address argument. Presumably that is more difficult than it sounds, given these clients are code gen'd.

neilcampbell:
> @neilcampbell Are we ok with covering support for remaining msgpack related abstractions in follow up PRs? I'm concerned with the size of this PR and would prefer having this one well reviewed merged and then having remaining bits in smaller chunks.

That seems ok to me.

crates/algokit_msgpack_ffi/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+    model_type: ModelType,
+    base64_str: &str,
+) -> Result<String, AlgoKitMsgPackError> {
+    let internal_type: InternalModelType = model_type.into();
+    Ok(decode_base64_msgpack_to_json(internal_type, base64_str)?)
+}
+
+#[ffi_func]
+pub fn list_supported_models_ffi() -> Vec<ModelType> {
+    algokit_msgpack::list_supported_models()
```
neilcampbell, line 136:
```suggestion
pub fn supported_models_ffi() -> Vec<ModelType> {
    algokit_msgpack::supported_models()
```

packages/typescript/algokit_msgpack/release.config.cjs
```diff
@@ -2,8 +2,7 @@ const releaseUtils = require("../../../utils/semantic-release.cjs");
 
 const config = releaseUtils.getConfig({
```
neilcampbell, line 3: Looks like this file was deleted from `algod_api`, however it shouldn't have.

scripts/build/languages/typescript.ts
```diff
@@ -68,5 +68,19 @@ export async function buildTypescript(crate: string) {
...
+    `packages/typescript/${crate}/pkg/${crate}_ffi.d.ts`,
+    `packages/typescript/${crate}/dist/index.d.ts`,
+  );
+
+  // Ensure the raw wasm blob is present next to the *.node.cjs bundle that
+  // Node (and Vitest when transpiling to CommonJS) will load via
+  //   readFileSync(path.join(__dirname, '<crate>_ffi_bg.wasm'))
+  // Rollup's CJS output does not inline the bytes, so we need to ship the
+  // file.  Copy it from the wasm-pack output directory into dist/.
+  fs.copyFileSync(
```
PatrickDinh, line 82: Can you please verify if this is still needed? I think after this [PR](https://github.com/algorandfoundation/algokit-core/pull/93), the wasm binary is always inline as base64 for CJS.

api/oas_templates/python/model_generic.mustache
```diff
@@ -394,6 +408,61 @@ class {{classname}}({{#parent}}{{{.}}}{{/parent}}{{^parent}}BaseModel{{/parent}}
...
+    @classmethod
+    def from_msgpack(cls, data: bytes) -> "Self":  # pragma: no cover – thin wrapper
+        """Create a new instance from Algorand MessagePack *data*.
+
+        The inverse of :pymeth:`to_msgpack`. Requires
+        ``algokit_msgpack`` to be importable.
+        """
+        if _ak_decode_msgpack is None or _AkModelType is None:
+            raise RuntimeError(
+                "algokit_msgpack is not available — install the extra binary wheel "
```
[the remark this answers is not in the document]
aorumbayev: Fixed

scripts/build/languages/typescript.ts
aorumbayev: @PatrickDinh verified that it works without the copy.

PR #106, opened 2025-05-25
fix: correct id raw naming to match encode raw convention

neilcampbell:
Renaming the method for generating a raw id (`id_raw`) to match the naming convention of raw encoding (`encode_raw`).


packages/swift/AlgoKitTransact/Frameworks/algokit_transact.xcframework/ios-arm64/Headers/algokit_transactFFI.h
PatrickDinh, line 1: Unrelated to this PR: should we remove these autogenerated files from Git?

PR #108, opened 2025-05-27
fix: changing to a build backend that correctly generated wheels with dependencies

CiottiGiorgio:
This fix changes the build backend used by poetry and the resulting wheel correctly specifies a METADATA file within the archive that contains the transitive dependencies of this package


api/oas_templates/python/pyproject.mustache
```diff
@@ -43,8 +43,8 @@ algokit-utils = "4.0.0"
 algokit-transact = { path = "../../../packages/python/algokit_transact" }
 
 [build-system]
-requires = ["setuptools"]
-build-backend = "setuptools.build_meta"
+requires = ["poetry-core>=1.0.0"]
```
aorumbayev, line 46: tried a fresh poetry init on latest v2, the version specifier it drops under default build system is:

```
requires = ["poetry-core>=2.0.0,<3.0.0"]
build-backend = "poetry.core.masonry.api"
```

Wonder if we should set the same specifier, considering poetry v1 is fairly old at this point

aorumbayev (approved):

PR #111, opened 2025-05-28
feat: assign fee to transaction

PatrickDinh:


crates/algokit_transact/src/tests.rs
```diff
@@ -129,3 +130,70 @@ fn test_estimate_transaction_size() {
...
+            min_fee: 1000,
+            extra_fee: None,
+            max_fee: None,
+        })
+        .unwrap();
+    assert_eq!(updated_transaction.header().fee, Some(1000));
+}
+
+#[test]
+fn test_extra_fee() {
```
neilcampbell, line 150: Might be worth updating one of these tests to ensure `assign_fee` isn't mutating the input txn.

crates/algokit_transact/src/transactions/mod.rs
```diff
@@ -46,6 +46,56 @@ pub enum Transaction {
...
+    pub fn header_mut(&mut self) -> &mut TransactionHeader {
+        match self {
+            Transaction::Payment(p) => &mut p.header,
+            Transaction::AssetTransfer(a) => &mut a.header,
+        }
+    }
+
+    pub fn assign_fee(&self, request: FeeParams) -> Result<Transaction, AlgoKitTransactError> {
+        let mut tx = self.clone();
+        let estimate_size = tx.estimate_size()?;
```
neilcampbell, line 73:
```suggestion
        let estimated_size = tx.estimate_size()?;
```

crates/algokit_transact/src/transactions/mod.rs
```diff
@@ -46,6 +46,56 @@ pub enum Transaction {
...
+        let mut estimate_fee = request.fee_per_byte * estimate_size as u64;
```
neilcampbell, line 74:
```suggestion
        let mut calculated_fee = request.fee_per_byte * estimate_size as u64;
```

crates/algokit_transact/src/transactions/mod.rs
neilcampbell: If the `fee_per_byte` is 0 we don't need to estimate the size of perform the calc.

crates/algokit_transact/src/transactions/mod.rs
```diff
@@ -46,6 +46,56 @@ pub enum Transaction {
...
+        }
+
+        if let Some(extra_fee) = request.extra_fee {
+            estimate_fee += extra_fee;
+        }
+
+        if let Some(max_fee) = request.max_fee {
+            if estimate_fee > max_fee {
+                return Err(AlgoKitTransactError::InputError(format!(
+                    "Transaction fee {} µALGO is greater than maxFee {} µALGO",
```
neilcampbell, line 86:
```suggestion
                    "Calculated transaction fee {} µALGO is greater than max fee {} µALGO",
```

crates/algokit_transact_ffi/src/lib.rs
```diff
@@ -162,7 +172,7 @@ pub struct Transaction {
     /// The sender of the transaction
     sender: Address,
 
-    fee: u64,
+    fee: Option<u64>,
```
neilcampbell, line 175: Question for the team: Are we ok with making fee Optional. In the case, it's not set it's omitted from transaction when sent.

It's probably worth building a txn with a None fee set, signing, encoding and sending using goal to ensure that it's a valid request.

crates/algokit_transact_ffi/src/lib.rs
```diff
@@ -522,6 +532,33 @@ pub fn get_algorand_constant(constant: AlgorandConstant) -> u64 {
...
+        })
+    }
+}
+
+#[ffi_func]
+pub fn assign_fee(
+    txn: &Transaction,
+    fee_params: FeeParams,
+) -> Result<Transaction, AlgoKitTransactError> {
+    let txn_internal: algokit_transact::Transaction = txn.clone().try_into()?;
```
neilcampbell, line 553: Do we need the clone? `txn_internal.assign_fee` performs a clone.

crates/algokit_transact_ffi/src/lib.rs
```diff
@@ -522,6 +532,33 @@ pub fn get_algorand_constant(constant: AlgorandConstant) -> u64 {
...
+    let fee_calculation_request_internal: algokit_transact::FeeParams =
```
neilcampbell, line 554:
```suggestion
    let fee_params_internal: algokit_transact::FeeParams =
```

crates/algokit_transact_ffi/src/lib.rs
```diff
@@ -522,6 +532,33 @@ pub fn get_algorand_constant(constant: AlgorandConstant) -> u64 {
...
+        fee_params.clone().try_into()?;
```
neilcampbell, line 555: Do we need clone for `fee_params`?

crates/algokit_transact/src/utils.rs
```diff
@@ -37,6 +37,10 @@ pub fn is_zero(n: &u64) -> bool {
     *n == 0u64
 }
 
+pub fn is_zero_opt(n: &Option<u64>) -> bool {
+    n.map_or(true, |v| v == 0)
```
neilcampbell, line 41:
```suggestion
    n.map_or(true, is_zero)
```

crates/algokit_transact/src/tests.rs
PatrickDinh: I'm wondering if we need to do this, or should we consider / make it a convention that functions shouldn't mutate the input data. I'm asking because if we add a test here, it makes sense to add similar tests everywhere we pass an object in as an param.

crates/algokit_transact/src/utils.rs
PatrickDinh: calling `is_zero` directly doesn't work because the value at this point already has type `u64`, not `Option<u64>` anymore. I updated it to `|n: u64| n == 0u64` to just to make sure the `0` is `u64`

crates/algokit_transact/src/tests.rs
joe-p: Isn't that conveyed by the signature though? We take in a `&self` and return `Transaction`, so the function can't mutate `self` and must return an owned value

crates/algokit_transact_ffi/src/lib.rs
joe-p: As discussed in today's meeting, optional fee field makes sense and is equivalent to 0 fee

crates/algokit_transact/src/transactions/mod.rs
neilcampbell: This is an interesting.
The main intention behind the suggestion was to produce a language agnostic error message.
Given we're sharing between both python and ts utils, we'll likely have be changing the message somewhere given that python will have a slightly different message to match the python semantics, so it's probably `max_fee`.

crates/algokit_transact/src/utils.rs
neilcampbell: Signature for is_zero is `is_zero(n: &u64) -> bool`, so it's operating a `u64` not an `Option<u6>`.

crates/algokit_transact/src/tests.rs
PatrickDinh: good point @joe-p

crates/algokit_transact/src/transactions/mod.rs
PatrickDinh: I see, I didn't notice the different between "maxFee" and "max fee". I agree that this should be updated to a language agnostic error message.

crates/algokit_transact_ffi/src/lib.rs
PatrickDinh: I updated this function to receive a transaction value instead of a reference => I also could remove this `clone`.

crates/algokit_transact/src/utils.rs
PatrickDinh: I found it. It wasn't the `Option` but the ref. I updated this function to
```
    n.as_ref().map_or(true, is_zero)
```

PR #119, opened 2025-06-03
spike: utils

joe-p:
Spike to see what a rust-based utils would be like. Focusing on Python and TS.


crates/algokit_utils_ffi/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+uniffi::setup_scaffolding!();
+
+#[derive(Debug, thiserror::Error, uniffi::Error)]
+pub enum ComposerError {
+    #[error("TransactionsError: {0}")]
+    TransactionsError(String),
+}
+
+#[derive(uniffi::Object)]
+pub struct Composer {
```
aorumbayev, line 16: Would the idea be to expose all main high level utils abstractions but excluding any user facing methods that would interact with API clients or any excessively stateful logic (deferring the implementation to the target language ffi binding package)? For example how would we implement `send` method on a `TransactionComposer` - i presume it wont be possible to handle it completely on rust side (given extra fact that api clients are handled separately per each target language) but is there a way to expose some sort of abstract stub that can then be implemented in target package (py/ts/...) by ourselves manually?

aorumbayev (commented):
It be great to also capture the spike (once its finalized) as an architecture diagram or a md document that highlights all user facing utils interfaces (based on current py vs ts interfaces that maintain feature parity) and lists down their public user facing methods/functions, as well as which of those would be exposed via ffi vs which would only have a stub and would have to be implemented manually in binding packages in target languages (referring to packages in packages folder).

PR #151, opened 2025-06-10
feat: add support for rekeyed account signing

neilcampbell:
Allows rekeyed accounts to sign by attaching a signer.


crates/algokit_transact/src/transactions/mod.rs
```diff
@@ -130,10 +132,16 @@ pub struct SignedTransaction {
     #[serde(rename = "txn")]
     pub transaction: Transaction,
 
-    /// The Ed25519 signature authorizing the transaction.
+    /// Optional Ed25519 signature authorizing the transaction.
     #[serde(rename = "sig")]
-    #[serde_as(as = "Bytes")]
-    pub signature: [u8; ALGORAND_SIGNATURE_BYTE_LENGTH],
+    #[serde_as(as = "Option<Bytes>")]
```
joe-p, line 137: I assume we switched to Option here because in the future there might be an lsig or msig?

crates/algokit_transact/src/transactions/mod.rs
```diff
@@ -130,10 +132,16 @@ pub struct SignedTransaction {
...
+    pub signature: Option<[u8; ALGORAND_SIGNATURE_BYTE_LENGTH]>,
+
+    /// Optional signer address applicable if the transaction sender is a rekeyed account.
+    #[serde(rename = "sgnr")]
+    #[serde(skip_serializing_if = "is_zero_addr_opt")]
+    #[serde(default)]
+    pub signer: Option<Address>,
```
joe-p, line 144: I'm wondering if `signer` or `auth_addr` is better. This is how it's defined in go-algorand:

```go
type SignedTxn struct {
	_struct struct{} `codec:",omitempty,omitemptyarray"`

	Sig      crypto.Signature   `codec:"sig"`
	Msig     crypto.MultisigSig `codec:"msig"`
	Lsig     LogicSig           `codec:"lsig"`
	Txn      Transaction        `codec:"txn"`
	AuthAddr basics.Address     `codec:"sgnr"`
}
```

Don't have a strong opinion just calling it out

joe-p (approved):
LGTM. Would be cursious to know if there's strong opinions on the name but we can always change later if we want.

crates/algokit_transact/src/transactions/mod.rs
neilcampbell: I'm also planning of using `auth_address` rather than `auth_addr`, as we use `transaction` instead of `txn` and `signature` instead of `sig`.

PR #163, opened 2025-06-18
feat(transact): keyreg txn support

joe-p:
## TODO

- [x] Wait for #165 and adopt that testing strategy in this PR
- [x] Implement validity checking once we decide on how we want to do that

## Summary

Entirely written by claude. Haven't yet reviewed anything, but wanted to commit and PR first so we can all see what it's first shot looked like.

  1. Explored codebase structure - Understood the existing transaction types (Payment, AssetTransfer) and how they're implemented
  2. Researched KeyReg transaction fields - Found comprehensive documentation about Algorand KeyReg transaction fields from official sources
  3. Implemented KeyReg transaction type - Added complete support including:
    - KeyRegistrationTransactionFields struct with all required fields:
        - vote_key (32-byte root participation public key)
      - selection_key (32-byte VRF public key)
      - state_proof_key (64-byte state proof key)
      - vote_first (first valid round)
      - vote_last (last valid round)
      - vote_key_dilution (key dilution parameter)
      - non_participation (optional non-participating flag)
    - Builder pattern support via KeyRegistrationTransactionBuilder
    - Integration with Transaction enum using the correct serde tag "keyreg"
    - FFI support in the algokit_transact_ffi crate
  4. Comprehensive tests covering:
    - Creating online key registration transactions
    - Updating (offline key registration transactions)
    - Deleting (non-participating key registration)
    - Calling in transaction groups with other transaction types
    - Encoding/decoding round-trip tests
    - Transaction ID generation
    - Fee calculation
    - Online/offline detection

  Key Features Implemented:

  - Online Registration: Full support for registering accounts online for consensus participation with all required participation keys
  - Offline Registration: Support for taking accounts offline (empty participation key fields)
  - Non-Participating: Support for marking accounts as non-reward earning
  - Canonical Fields: Used exact field names and types from Algorand documentation (votekey, selkey, sprfkey, votefst, votelst, votekd, nonpart)
  - Proper Serialization: Uses MessagePack with correct field naming and optional field handling
  - Transaction Grouping: KeyReg transactions can be grouped with other transaction types
  - Builder Pattern: Consistent with existing transaction builders for easy construction

  The implementation follows the existing patterns in the codebase and includes comprehensive tests as requested. The KeyReg transaction type is now fully functional and ready to use in applications that need to manage Algorand account
  participation in consensus.


crates/algokit_transact/src/transactions/mod.rs
```diff
@@ -116,6 +121,12 @@ impl AssetTransferTransactionBuilder {
     }
 }
 
+impl KeyRegistrationTransactionBuilder {
```
joe-p, line 124: I noticed we have all the builders defined here. Probably makes sense to move them into their respective files?

crates/algokit_transact/src/transactions/key_registration.rs
```diff
@@ -0,0 +1,110 @@
...
+    /// Mark account as non-reward earning.
+    /// Optional field that can be used with both online and offline registration.
+    #[serde(rename = "nonpart")]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(default)]
+    #[builder(default)]
+    pub non_participation: Option<bool>,
+}
+
+impl KeyRegistrationTransactionFields {
```
joe-p, line 87: I like this approach. Perhaps we have a `ValidTransaction` trait that has a `valid` function with a default implementation that does checks on the header and the calls a `valid_for_type` function that is called internally to do specific checks. 

As an example (untested, may have errors):

```rust
trait ValidTransaction {
    fn valid_for_type(&self) Result<(), String>;

    fn valid(&self) Result<(), String> {
        // Do checks on the header fields
        self.valid_for_type()
    }
}

impl ValidTransaction for KeyRegistrationTransaction {
    fn is_valid_for_type(&self) Result<(), String> {
        self.valid_offline()?
        self.valid_online()
    }
}
```

crates/algokit_transact/src/transactions/key_registration.rs
```diff
@@ -0,0 +1,110 @@
...
+            && self.selection_key.is_some()
+            && self.state_proof_key.is_some()
+            && self.vote_first.is_some()
+            && self.vote_last.is_some()
+            && self.vote_key_dilution.is_some()
+    }
+
+    /// Returns true if this is an offline key registration transaction.
+    /// Offline transactions have no participation key fields set.
+    pub fn is_offline(&self) -> bool {
```
joe-p, line 101: Should also have a specific check for non-part

PatrickDinh (commented):
oops, I approved by mistake.

crates/algokit_transact_ffi/src/lib.rs
```diff
@@ -436,11 +462,30 @@ fn byte32_to_bytebuf(b32: Byte32) -> ByteBuf {
     ByteBuf::from(b32.to_vec())
 }
 
+fn bytebuf_to_32_bytes(buf: &ByteBuf) -> Result<[u8; 32], AlgoKitTransactError> {
```
PatrickDinh, line 465: I think this function duplicates the existing bytebuf_to_byte32 function

crates/algokit_transact/src/transactions/key_registration.rs
neilcampbell: @PatrickDinh Has done up an implementation of this in his PR https://github.com/algorandfoundation/algokit-core/pull/161/files
The naming is slightly different (`trait Validate` with `validate`) and there isn't any header validation as the header type mostly prevents invalid data being set, however that can be added like you have above very easily.

neilcampbell (commented):
@joe-p Shout once you've updated based on the new conventions and we'll review.

crates/algokit_transact/src/transactions/key_registration.rs
aorumbayev: @neilcampbell added trait Validate impl, ready for re review

crates/algokit_transact_ffi/src/lib.rs
aorumbayev: @PatrickDinh combine byte32 and byte64 into a single helper method hence bytebuf_to_32_bytes bytebuf_to_byte32 and bytebuf_to_byte64 are removed now

crates/algokit_transact/src/transactions/key_registration.rs
```diff
@@ -0,0 +1,434 @@
...
+                "Key registration validation failed: {}",
+                errors.join("\n")
+            ))
+        })?;
+        Ok(Transaction::KeyRegistration(d))
+    }
+}
+
+impl Validate for KeyRegistrationTransactionFields {
+    fn validate(&self) -> Result<(), Vec<String>> {
```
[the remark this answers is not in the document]
neilcampbell: Also asset config uses a different approach which doesn't include `is_online` etc. I think we should probably keep consistent.

crates/algokit_transact_ffi/src/transactions/keyreg.rs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            return Err(Self::Error::DecodingError(
+                "Key Registration data missing".to_string(),
+            ));
+        }
+
+        let data = tx.clone().key_registration.unwrap();
+        let header: algokit_transact::TransactionHeader = tx.try_into()?;
+
+        Ok(Self {
```
neilcampbell, line 69: This should also have a call to the validator like asset config.

neilcampbell (approved):

PR #164, opened 2025-06-19
feat: consolidating the ASA transfer transaction

CiottiGiorgio:


CiottiGiorgio:
Good checkpoint w.r.t. AK-373

PR #165, opened 2025-06-20
feat: Add support for application call transactions

neilcampbell:
TODO
- [ ] Add additional data validation when crafting.


crates/algokit_transact/src/test_utils/mod.rs
```diff
@@ -261,6 +481,42 @@ impl TestDataMother {
         TransactionTestData::new(transaction, signing_private_key)
     }
 
+    pub fn application_create() -> TransactionTestData {
+        let signing_private_key: Byte32 = [
+            2, 205, 103, 33, 67, 14, 82, 196, 115, 196, 206, 254, 50, 110, 63, 182, 149, 229, 184,
+            216, 93, 11, 13, 99, 69, 213, 218, 165, 134, 118, 47, 44,
+        ];
```
PatrickDinh, line 488: nitpick: let's create a constant for this private_key and reuse it across all the functions

crates/algokit_transact/src/transactions/application_call.rs
```diff
@@ -0,0 +1,331 @@
...
+                .map(|box_ref| {
+                    let app_id_index = if box_ref.app_id == 0 || box_ref.app_id == fields.app_id {
+                        // App ID 0 and the transaction's own app ID remain as 0 (current application)
+                        0
+                    } else {
+                        // Find position in app_references and add 1 (1-based indexing)
+                        app_references
+                            .iter()
+                            .position(|&id| id == box_ref.app_id)
+                            .map(|pos| (pos + 1) as u64) // App references start from index 1; index 0 is ithe current application ID.
```
PatrickDinh, line 256:
```suggestion
                            .map(|pos| (pos + 1) as u64) // App references start from index 1; index 0 is the current application ID.
```

crates/algokit_transact/src/transactions/application_call.rs
```diff
@@ -0,0 +1,331 @@
...
+    // Transform box references if present
+    if let Some(ref box_references) = fields.box_references {
+        if !box_references.is_empty() {
+            let app_references = fields.app_references.as_deref().unwrap_or(&[]);
+
+            let box_references = box_references
+                .iter()
+                .map(|box_ref| {
+                    let app_id_index = if box_ref.app_id == 0 || box_ref.app_id == fields.app_id {
+                        // App ID 0 and the transaction's own app ID remain as 0 (current application)
```
PatrickDinh, line 249: I think the comment can be a bit clearer
```suggestion
                        // Return 0 if the App ID is 0 or the App ID is the same with the current application ID
```
something along this line, what do you think?

PatrickDinh (approved):

PR #182, opened 2025-06-27
chore: integration test for sending and waiting for asset transfers

CiottiGiorgio:


CiottiGiorgio:
I will start over with this task once the blocking changes on the API client are merged.

PR #185, opened 2025-06-30
feat: algorand fixture

PatrickDinh:
# Proposed changes

1. Consolidated test infrastructure: Removed standalone algod_client_tests crate and moved testing utilities into algokit_utils
2. Enhanced algokit_utils crate: Added new modules for client management (client_manager.rs), network configuration (network_client.rs), and Algorand client abstraction (algorand_client.rs)
3. Expanded testing framework: Added comprehensive testing utilities including localnet management, account helpers, and fixture support under src/testing/
4. Reorganized integration tests: Moved and organized tests into structured modules (tests/algod/, tests/composer/) with improved test organization
5. Updated CI workflow: Renamed algod_client_tests.yml to rust_tests.yml and expanded scope to run all workspace tests with enhanced debugging
6. Added dependencies: Integrated logging (log), environment configuration (dotenvy), async utilities (tokio), and cryptographic dependencies (moved from now removed algod transact tests used in mnemonic to pkey transformation)
7. Improved error handling: Enhanced localnet startup robustness and client configuration management


crates/algokit_utils/src/testing/fixture.rs
```diff
@@ -0,0 +1,165 @@
...
+pub struct TransactionResult {
+    pub transaction: Transaction,
+    pub tx_id: String,
+    pub signed_bytes: Vec<u8>,
+}
+
+// TODO: We will need to implement account signer to replace this struct
+// It is only here initially to get the tests working
+#[derive(Clone)]
+pub struct TestAccountSigner {
```
PatrickDinh, line 34: I think we need a better way to construct the AccountSigner. Ideally, we should create a trait 
```
trait TransactionSigner {
  fn sign_transaction(
        &self,
        transaction: &Transaction,
    ) -> Result<Vec<u8>, Box<dyn std::error::Error + Send + Sync>>
}
```
and the `TestAccount` struct should implement this trait. 
What do you think @aorumbayev? If you think that it's a good idea, where do you think that we should put it? It needs to be somewhere that the `algod_client_tests` can reference it.

crates/algokit_utils/tests/payment.rs
```diff
@@ -0,0 +1,57 @@
+use algokit_utils::testing::*;
+use algokit_utils::{CommonParams, PaymentParams};
+
+#[tokio::test]
+async fn test_basic_payment_transaction() {
+    // ARRANGE - Create fixture and new scope
```
neilcampbell, line 6: I don't really think the arrange/act/assert comments are needed.

crates/algokit_utils/tests/payment.rs
```diff
@@ -0,0 +1,57 @@
...
+        algokit_transact::Transaction::Payment(payment_fields) => {
+            assert_eq!(
+                payment_fields.amount, 500_000,
+                "Payment amount should be 500_000 microALGOs"
+            );
+        }
+        _ => panic!("Transaction should be a payment transaction"),
+    }
+
+    println!("✓ Basic payment test completed successfully");
```
neilcampbell, line 56: Feels unnecessary

crates/algokit_transact/src/test_utils/asset_config.rs
```diff
@@ -0,0 +1,114 @@
+use crate::{Address, AssetConfigTransactionBuilder, Byte32, TransactionHeaderBuilder};
+use base64::{prelude::BASE64_STANDARD, Engine};
+
+pub struct AssetConfigTransactionMother {}
+
+impl AssetConfigTransactionMother {
```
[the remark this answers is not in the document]
aorumbayev: @neilcampbell branch should be up to date against latest main, can you confirm its no longer an issue? Otherwise could you clarify what you were referring to here as i may be missing some context here after taking over the pr from @PatrickDinh

crates/algokit_utils/src/lib.rs
```diff
@@ -276,7 +315,100 @@ impl Composer {
...
+        let mut current_round = start_round;
+
+        while current_round < start_round + max_rounds {
+            let _ = match self
+                .algod_client
+                .pending_transaction_information(tx_id, Some("msgpack"))
+                .await
+            {
+                Ok(response) => {
+                    // TODO: handle pool error
```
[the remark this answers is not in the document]
aorumbayev: @PatrickDinh i believe this was referring to handling of the pool error from pending txn response? I've added a handling logic for it lmk if this look ok

crates/algokit_utils/src/lib.rs
```diff
@@ -276,7 +315,100 @@ impl Composer {
...
+                    if response.confirmed_round.is_some() {
+                        return Ok(response);
+                    }
+                }
+                Err(_) => {
+                    // TODO: only trigger this if the error is URLTokenBaseHTTPError
```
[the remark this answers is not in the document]
aorumbayev: @neilcampbell this seems to be a todo left by @joe-p . I've checked the URLTokenBaseHTTPError in algosdk, and it seems like it was re triggering the retry on all 4xx status codes that URLTokenBaseHTTPError can throw, i've tweaked it to handle specifically the 404 which i believe might be more precise in this case rather than retrying on all 4xx codes. Lmk if this looks ok to you @neilcampbell

aorumbayev:
@neilcampbell @PatrickDinh the last commit also tweaks the structure mirroring the utils-py's domain based structure. I feel like setting it up earlier than later will make things easier as new abstractions are being added, and we don't want to have entire utils rust implemented in a single lib.rs file as it will make navigating within the project complicated (both for humans and llms)

crates/algokit_utils/tests/algod/transaction_params.rs
PatrickDinh, line 1: I'm not sure if these tests are needed. They should be covered by any transaction integration tests.

crates/algokit_utils/tests/transactions/composer.rs
```diff
@@ -0,0 +1,61 @@
+use algokit_utils::testing::*;
+use algokit_utils::{CommonParams, PaymentParams};
+
+use crate::common::init_test_logging;
+
+#[tokio::test]
+async fn test_basic_payment_transaction() {
```
PatrickDinh, line 7: I think transaction type tests should still be in separated files because this file will get long if we add more tests.

crates/algokit_utils/tests/algod/transaction_params.rs
aorumbayev: I'll keep it there for now, the goal was to have a test file per endpoint, and given this being used as a base in many other integration tests i think its still useful to have it. If we ever hit any scenario where something is wrong specifically with this endpoint we still can benefit from this test giving ability to quickly debug this specific endpoint in isolation.

crates/algokit_utils/src/network_client.rs
```diff
@@ -0,0 +1,79 @@
...
+impl AlgoClientConfig {
+    pub fn new(server: String) -> Self {
+        Self {
+            server,
+            port: None,
+            token: None,
+        }
+    }
+
+    pub fn with_port(mut self, port: u16) -> Self {
```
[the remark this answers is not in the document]
aorumbayev: Agree, removed

PR #187, opened 2025-07-01
feat: multisignature accounts and transactions

CiottiGiorgio:
While introducing multisignature primitives, it's useful to decouple the concept of a (single signature) account to an address. Addresses are what allows composability with single sigs, multi sigs, logic sigs.

At this point, the proposed abstraction is ready for review.

Items that are still being worked on are:
- testing
- ffi code
- more refined msig primitives


CiottiGiorgio:
How the abstractions are exposed through FFI is probably not the most ergonomic at the moment, but further tasks can refine that, especially when using working on utils.

crates/algokit_transact/src/transactions/application_call.rs
```diff
@@ -190,7 +190,7 @@ pub struct ApplicationCallTransactionFields {
     #[serde(skip_serializing_if = "is_empty_vec_opt")]
     #[serde(default)]
     #[builder(default)]
-    pub account_references: Option<Vec<Address>>,
+    pub account_references: Option<Vec<Account>>,
```
neilcampbell, line 193: Shouldn't the type stay as `Option<Vec<Address>>`?

crates/algokit_transact/src/transactions/asset_config.rs
```diff
@@ -164,7 +164,7 @@ pub struct AssetConfigTransactionFields {
     ///
     /// If not set or set to the Zero address the asset becomes permanently immutable.
     #[builder(default)]
-    pub manager: Option<Address>,
+    pub manager: Option<Account>,
```
neilcampbell, line 167: Curious why the type was changed? A manager could be a multisig? Applies to all of these.

crates/algokit_transact_ffi/src/lib.rs
```diff
@@ -155,11 +243,11 @@ pub struct FeeParams {
 
 #[ffi_record]
 pub struct PaymentTransactionFields {
-    receiver: Address,
+    receiver: Account,
```
neilcampbell, line 246: I'm wondering if we make these strings?
I think it's worth putting the thought into getting the FFI layer supporting multisigs as well.

CiottiGiorgio: I'll try this change with Strings. Since uniffi does not support unnamed fields in a struct, we'd need to have `Address` struct have `addr` field, resulting in an account having an address as `account.address.addr` which is not very ergonomic

CiottiGiorgio: This should be addressed now

packages/python/algokit_transact/tests/transaction_asserts.py
```diff
@@ -26,6 +26,38 @@ def assert_example(test_data: TransactionTestData):
...
+def assert_multisig_example(test_data: TransactionTestData):
+    """A multisig example of forming a transaction and signing it"""
+    single_sig = test_data.signing_private_key.sign(
+        encode_transaction(test_data.transaction)
+    ).signature
+
+    # FIXME: Since we don't yet expose a way to create a MultisigSignature, we create it manually
+    #  with an empty address which shouldn't impact the encoding.
+    multisig_signature = MultisigSignature(
+        address="",
```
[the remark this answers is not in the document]
CiottiGiorgio: This will be part of following tickets

crates/algokit_transact_ffi/src/transactions/application_call.rs
```diff
@@ -54,7 +54,7 @@ pub struct ApplicationCallTransactionFields {
 
     /// List of accounts in addition to the sender that may be accessed
     /// from the application's approval program and clear state program.
-    account_references: Option<Vec<Address>>,
+    account_references: Option<Vec<Account>>,
```
neilcampbell, line 57: Should these be string now, as it's valid to use a multisig address?

CiottiGiorgio: Yes definitely, logisigs too

joe-p:
Not a blocker, but I think we should think about the naming for `Account`. It seems weird to me that `Account` is reserved specifically a single keypair account but a msig and lsig are also both accounts. Looking at our existing tooling/docs we typically refer to single keypair accounts as "standard" accounts, so maybe we call it `StandardAccount`? This would also help us separate from HD accounts which we will eventually support in core.

crates/algokit_transact/src/tests.rs
```diff
@@ -120,32 +123,41 @@ fn test_asset_opt_in_signed_transaction_encoding() {
...
         290,
-        Some(AddressMother::address().clone()),
+        Some(AccountMother::account().clone()),
     );
 }
 
 #[test]
 fn test_zero_address() {
-    let addr = AddressMother::zero_address();
+    let acct = AccountMother::zero_address();
```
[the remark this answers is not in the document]
CiottiGiorgio: I agree it makes sense when you think about it but it's weird to write it this way.
I'll change to `zero_address_account` because I think it's a better name but feel free to change it again to something else in the future if you find an even better name for it

neilcampbell (approved):

PR #189, opened 2025-07-04
feat(utils): Adds key registration support

aorumbayev:
Adds support for online, offline, and non-participation key
registration transactions to the composer.

This enhancement allows users to manage their account's
participation in the consensus protocol. Includes relevant
parameter structs and composer methods.

Also adds integration tests to verify the behavior of the
key registration transactions.


crates/algokit_utils/tests/transactions/composer/key_registration.rs
```diff
@@ -0,0 +1,497 @@
...
+            vote_key.to_vec(),
+            "Vote participation key should match submitted key"
+        );
+    } else {
+        panic!("Account should have participation information after online key registration");
+    }
+}
+
+#[tokio::test]
+async fn test_permanent_offline_key_registration_transaction() {
```
neilcampbell, line 326: This is technically the same as non participation key registration. Probably makes sense to fold any useful logic into that test.

crates/algokit_utils/src/transactions/composer.rs
```diff
@@ -731,6 +837,94 @@ mod tests {
         assert!(composer.built_group().is_some());
     }
 
+    #[tokio::test]
+    async fn test_build_online_key_registration() {
```
PatrickDinh, line 841: Do we need to test `composer.build()` when we have integration tests for `composer.send()` as well? I think the integration tests cover these tests.

aorumbayev: @neilcampbell whats your opinion on this, i agree with Pat but on the other hand they still feel like unit tests due to only relying on the get suggested params call, while integration tests we got a full fledged e2e tests that commit state to ledger. Maybe we use dummy suggested params instead and turn them into proper unit tests but still keep it separately as they can allow us detect build specific logic issues faster than a full blown localnet based test

neilcampbell (approved):

PR #190, opened 2025-07-04
chore: added integration test for sending and waiting for asset transfers

CiottiGiorgio:
The first FIXME could perhaps stay in the code until the asset configs are merged in the composer and be changed later, or we could wait for it in this PR. Depending on what the reviewer think is best that's what we'll do.

It's not clear to me at this moment how to solve the TODO.

### DIFF
--- a/crates/algokit_utils/tests/transactions/composer/asset_transfer.rs
+++ b/crates/algokit_utils/tests/transactions/composer/asset_transfer.rs
@@ -1,6 +1,6 @@
 use crate::common::init_test_logging;
 use algokit_utils::testing::*;
-use algokit_utils::transactions::composer::{AssetOptInParams, AssetTransferParams, SendParams};
+use algokit_utils::transactions::composer::{AssetOptInParams, AssetTransferParams};
 use algokit_utils::{AssetCreateParams, CommonParams};
 use std::sync::Arc;
 
@@ -14,16 +14,21 @@ async fn test_asset_transfer_transaction() {
         .new_scope()
         .await
         .expect("Failed to create new scope");
+
     let context = fixture.context().expect("Failed to get context");
 
-    let mut asset_create_composer = context.composer.clone();
-    let asset_creator = context.test_account.clone();
-    let asset_creator_address = asset_creator.account().unwrap().address();
-    asset_create_composer
+    let mut composer = context.composer.clone();
+
+    let asset_creator_address = context
+        .test_account
+        .account()
+        .expect("Failed to get sender account")
+        .address();
+
+    composer
         .add_asset_create(AssetCreateParams {
             common_params: CommonParams {
                 sender: asset_creator_address.clone(),
-                signer: Some(Arc::new(asset_creator)),
                 ..Default::default()
             },
             total: 10,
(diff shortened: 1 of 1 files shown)


crates/algokit_utils/tests/transactions/composer/asset_transfer.rs
```diff
@@ -0,0 +1,109 @@
...
+    let signed_asa_create_txn = asa_creator.sign_transaction(&asa_create_txn).unwrap();
+
+    let asa_create_id = algod_client
+        .raw_transaction(signed_asa_create_txn)
+        .await
+        .expect("Failed to send asset creation transaction");
+    let asa_create_result = algod_client
+        .pending_transaction_information(asa_create_id.tx_id.as_str(), Some(Msgpack))
+        .await
+        .expect("Failed to get pending transaction info");
```
PatrickDinh, line 58: I think sending the asset create txn without the composer is probably fine.

However, you should use the `wait_for_confirmation` logic from the composer because there is no guarantee that the transaction will success straight away.

crates/algokit_utils/tests/transactions/composer/asset_transfer.rs
```diff
@@ -0,0 +1,109 @@
...
+            },
+            asset_id: asa_create_result.asset_index.unwrap(),
+            receiver: asa_creator.address().unwrap(),
+            amount: 0,
+        })
+        .expect("Failed to add asset transfer");
+
+    let result = composer.send().await.expect("Failed to send transaction");
+
+    // FIXME: How to access the second transaction?
```
PatrickDinh, line 86: you're right here, currently the composer only returns the 1st transaction in a group. It is a missing featured compared to the existing utils-ts. It is a feature that we should add.

crates/algokit_utils/tests/transactions/composer/asset_transfer.rs
CiottiGiorgio: Good point

CiottiGiorgio:
The first FIXME is addressed by #193

CiottiGiorgio:
@PatrickDinh this is ready for a second pass if you would like to do so. If not, I'll go ahead and merge.

crates/algokit_utils/tests/transactions/composer/asset_transfer.rs
PatrickDinh: maybe I missed something, I thought that you could set the signer for each individual transaction. This is the common params
```
#[derive(Debug, Default, Clone)]
pub struct CommonParams {
    pub sender: Address,
    #[debug(skip)]
    pub signer: Option<Arc<dyn TxnSigner>>,
    pub rekey_to: Option<Address>,
    pub note: Option<Vec<u8>>,
    pub lease: Option<[u8; 32]>,
    pub static_fee: Option<u64>,
    pub extra_fee: Option<u64>,
    pub max_fee: Option<u64>,
    pub validity_window: Option<u64>,
    pub first_valid_round: Option<u64>,
    pub last_valid_round: Option<u64>,
}
```
`pub signer: Option<Arc<dyn TxnSigner>>,` should let you to pass a signer in.

CiottiGiorgio: I actually ended up building a `new_scope()` twice and using different context, each aware of their separate signer

PR #191, opened 2025-07-06
feat: add validation and integrate app call transaction

neilcampbell:
- Add validation for app call transactions


crates/algokit_transact/src/transactions/application_call.rs
```diff
@@ -4,13 +4,41 @@
...
-use crate::{Transaction, TransactionHeader};
+use crate::Transaction;
 use derive_builder::Builder;
 use serde::{Deserialize, Serialize};
 use serde_repr::{Deserialize_repr, Serialize_repr};
 use serde_with::{serde_as, skip_serializing_none, Bytes};
 
+// Application program size constraints (based on Algorand protocol)
+const MAX_EXTRA_PROGRAM_PAGES: u64 = 3; // Maximum extra program pages
+const PROGRAM_PAGE_SIZE: usize = 2048; // Size in bytes of each extra program page
```
PatrickDinh, line 18: nitpick: I think the constant names are good enough, don't need comments here

crates/algokit_transact/src/transactions/application_call.rs
```diff
@@ -4,13 +4,41 @@
...
+
+// Application reference limits
+const MAX_APP_ARGS: usize = 16; // Maximum number of args allowed
+const MAX_ARGS_SIZE: usize = 2048; // Maximum size in bytes of all args combined
```
PatrickDinh, line 22: nitpick: same as above, I think the constant names are good enough, don't need comments here

crates/algokit_transact/src/transactions/application_call.rs
```diff
@@ -334,10 +356,341 @@ where
...
+
+    /// Validates fields specific to application update.
+    pub fn validate_for_update(&self) -> Result<(), Vec<TransactionValidationError>> {
+        let mut errors = Vec::new();
+
+        self.validate_app_id_not_zero(&mut errors);
+        self.validate_programs_required(&mut errors);
+        // We can't validate the extra program pages on update, as we don't know what the initial create value was.
+        self.validate_immutable_fields_not_set(&mut errors);
+        self.validate_common_fields(&mut errors);
```
PatrickDinh, line 507: wondering it we could pushing new errors into the variable `errors` instead of mutating it?
It could be a noob question as I'm new to Rust

crates/algokit_utils/src/transactions/application_call.rs
```diff
@@ -0,0 +1,142 @@
...
+    pub app_id: u64,
+    /// Defines what additional actions occur with the transaction.
+    pub on_complete: OnApplicationComplete,
+    /// Transaction specific arguments available in the application's
+    /// approval program and clear state program.
+    pub args: Option<Vec<Vec<u8>>>,
+    /// List of accounts in addition to the sender that may be accessed
+    /// from the application's approval program and clear state program.
+    pub account_references: Option<Vec<Address>>,
+    /// List of applications in addition to the application ID that may be called
```
PatrickDinh, line 20:
```suggestion
    /// List of applications in addition to the current application that may be called
```
I think this is easier to understand

crates/algokit_utils/src/transactions/application_call.rs
```diff
@@ -0,0 +1,142 @@
...
+    /// Currently, the maximum value is 3.
+    /// This cannot be changed after creation.
+    pub extra_program_pages: Option<u64>,
+    /// Transaction specific arguments available in the application's
+    /// approval program and clear state program.
+    pub args: Option<Vec<Vec<u8>>>,
+    /// List of accounts in addition to the sender that may be accessed
+    /// from the application's approval program and clear state program.
+    pub account_references: Option<Vec<Address>>,
+    /// List of applications in addition to the application ID that may be called
```
PatrickDinh, line 73:
```suggestion
    /// List of applications in addition to the current application that may be called
```

crates/algokit_utils/src/transactions/application_call.rs
```diff
@@ -0,0 +1,142 @@
...
+    /// Clear state programs cannot reject the transaction.
+    /// Only required for application creation and update transactions.
+    pub clear_state_program: Vec<u8>,
+    /// Transaction specific arguments available in the application's
+    /// approval program and clear state program.
+    pub args: Option<Vec<Vec<u8>>>,
+    /// List of accounts in addition to the sender that may be accessed
+    /// from the application's approval program and clear state program.
+    pub account_references: Option<Vec<Address>>,
+    /// List of applications in addition to the application ID that may be called
```
PatrickDinh, line 132:
```suggestion
    /// List of applications in addition to the current application that may be called
```

crates/algokit_utils/src/transactions/application_call.rs
```diff
@@ -0,0 +1,142 @@
...
+/// Parameters for application call transactions.
+#[derive(Debug, Clone)]
+pub struct ApplicationCallParams {
+    pub common_params: CommonParams,
+    /// ID of the application being called.
+    ///
+    /// Set this to 0 to indicate an application creation call.
+    pub app_id: u64,
+    /// Defines what additional actions occur with the transaction.
+    pub on_complete: OnApplicationComplete,
```
PatrickDinh, line 13: should we make a new enum so that the user only can call this with valid on complete action for app call? same comment applies for create/delete too. 
I'm thinking about something similar to `utils-ts`. However, this is Rust and it could be a lot of work involved thought so it may not worth it.

crates/algokit_utils/tests/transactions/composer/application_call.rs
```diff
@@ -0,0 +1,363 @@
...
+    match transaction {
+        algokit_transact::Transaction::ApplicationCall(application_call_fields) => {
+            assert_eq!(
+                application_call_fields.app_id, app_id,
+                "Application ID should match"
+            );
+            assert_eq!(
+                application_call_fields.on_complete,
+                OnApplicationComplete::DeleteApplication,
+                "Clear state program should be DeleteApplication"
```
PatrickDinh, line 229:
```suggestion
                "On Complete should be DeleteApplication"
```

crates/algokit_transact/src/transactions/application_call.rs
neilcampbell: Line 18, I think it's good to clarify that it's bytes, so will simplify the comment.

PR #197, opened 2025-07-10
Fix custom signer for transaction

PatrickDinh:
When a signer is set in the transaction params, it should be used for signing. In this PR, I also added logic to batch signing (see `gather_signatures` function for efficiency.

### DIFF
--- a/crates/algokit_utils/src/transactions/composer.rs
+++ b/crates/algokit_utils/src/transactions/composer.rs
@@ -10,7 +10,7 @@ use algokit_transact::{
     Transactions,
 };
 use derive_more::Debug;
-use std::sync::Arc;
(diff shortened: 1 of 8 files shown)


crates/algokit_utils/src/transactions/composer.rs
```diff
@@ -40,6 +40,18 @@ pub enum ComposerError {
     PoolError(String),
 }
 
+#[derive(Clone)]
+pub struct TransactionWithSigner {
+    pub transaction: Transaction,
+    pub signer: Arc<dyn TxnSigner>,
+}
+
+impl TransactionWithSigner {
```
PatrickDinh, line 49: maybe we don't need this

PR #200, opened 2025-07-10
Pr feedback

PatrickDinh:
Fixes #

## Proposed Changes

-
-
-


crates/algokit_transact/src/test_utils/asset_freeze.rs
```diff
@@ -0,0 +1,81 @@
+use crate::{Address, AssetFreezeTransactionBuilder, Byte32, TransactionHeaderBuilder};
+use base64::{prelude::BASE64_STANDARD, Engine};
+
+pub struct AssetFreezeTransactionMother {}
```
PatrickDinh, line 4: test data builder for asset freeze are refactored to another file, similar to https://github.com/algorandfoundation/algokit-core/blob/main/crates/algokit_transact/src/test_utils/asset_config.rs

crates/algokit_transact/src/test_utils/asset_freeze.rs
```diff
@@ -0,0 +1,81 @@
...
+                    .unwrap(),
+            )
+            .asset_id(1707148495)
+            .freeze_target(freeze_address)
+            .frozen(true)
+            .to_owned()
+    }
+
+    pub fn asset_unfreeze() -> AssetFreezeTransactionBuilder {
+        // testnet-LZ2ODDAT4ATAVJUEQW34DIKMPCMBXCCHOSIYKMWGBPEVNHLSEV2A
```
PatrickDinh, line 50: An asset unfreeze transaction is built with real data from testnet. This allows us to confirm that our encoding is correct.

crates/algokit_transact/src/test_utils/mod.rs
```diff
@@ -708,4 +623,13 @@ mod tests {
             String::from("2XFGVOHMFYLAWBHOSIOI67PBT5LDRHBTD3VLX5EYBDTFNVKMCJIA")
         );
     }
+
+    #[test]
+    fn test_asset_unfreeze_snapshot() {
```
PatrickDinh, line 628: Test snapshot to make sure that the encoding logic is correct.

crates/algokit_transact/src/tests.rs
```diff
@@ -410,167 +410,3 @@ fn test_signed_transaction_group_encoding() {
         assert_eq!(decoded_signed_tx, signed_grouped_tx);
     }
 }
-
-#[test]
-fn test_asset_freeze_transaction_encoding() {
```
PatrickDinh, line 415: We don't need this test because the encoding logic is ready verified by the snapshot test. For decoding, I am not sure if we need tests for each transaction type because most of the heavy lifting is done by serde.

crates/algokit_transact/src/tests.rs
```diff
@@ -410,167 +410,3 @@ fn test_signed_transaction_group_encoding() {
...
-    let unfreeze_tx = TransactionMother::asset_unfreeze().build().unwrap();
-    let unfreeze_id = unfreeze_tx.id().unwrap();
-    assert_eq!(unfreeze_id.len(), 52);
-
-    // Verify freeze and unfreeze have different IDs (different frozen value)
-    assert_ne!(freeze_id, unfreeze_id);
-}
-
-#[test]
-fn test_asset_freeze_required_fields() {
```
PatrickDinh, line 521: I think this isn't needed, it's testing the builder which is a 3rd party library. I'd trust that they have their own test suite & we don't need to test their logic.

crates/algokit_transact/src/tests.rs
```diff
@@ -410,167 +410,3 @@ fn test_signed_transaction_group_encoding() {
...
-        .header(TransactionHeaderMother::simple_testnet().build().unwrap())
-        .asset_id(12345)
-        .frozen(true)
-        .build();
-
-    assert!(result.is_err());
-}
-
-#[test]
-fn test_asset_freeze_serialization_fields() {
```
PatrickDinh, line 543: Feels like this is testing serde

crates/algokit_transact/src/transactions/asset_freeze.rs
```diff
@@ -46,9 +46,9 @@ pub struct AssetFreezeTransactionFields {
     /// `false` to unfreeze the asset holdings (allow transfers).
     #[serde(rename = "afrz")]
     #[serde(default)]
-    #[serde(skip_serializing_if = "is_false_opt")]
+    #[serde(skip_serializing_if = "std::ops::Not::not")]
     #[builder(default)]
-    pub frozen: Option<bool>,
+    pub frozen: bool,
```
PatrickDinh, line 51: Sorry I missed your comment earlier this week. The reason is was done this way is the frozen field is a required field for asset freeze transaction. However, only for encoding purposed, it's skipped if the value is False.

crates/algokit_transact/src/test_utils/mod.rs
```diff
@@ -554,17 +474,12 @@ fn normalise_json(value: serde_json::Value) -> serde_json::Value {
...
-    const BOOLEAN_FIELDS_TO_KEEP: &[&str] = &["frozen"];
-
     match value {
         serde_json::Value::Object(map) => serde_json::Value::Object(
             map.into_iter()
                 .filter(|(k, v)| {
                     !(v.is_null()
-                        || v.is_boolean()
-                            && v.as_bool() == Some(false)
-                            && !BOOLEAN_FIELDS_TO_KEEP.contains(&k.to_case(Case::Snake).as_str())
```
PatrickDinh, line 567: This is not needed, the `defaultReviver` in TypeScript and Python should set it to false if undefined.

PR #213, opened 2025-07-21
Add ABI types

PatrickDinh:
## Proposed Changes

- Add ABI collection and primitive types
  - support encode/decode
  - support parsing strings to ABI types


crates/algokit_abi/README.md
PatrickDinh, line 1: I leave it empty until we have more in this crate

aorumbayev (approved):
would need to add some rustdocs and a final readme after we get the method sig pr merged here other than that approving

crates/algokit_transact/Cargo.toml
```diff
@@ -20,8 +20,9 @@ serde_bytes = "0.11.12"
 serde_json = "1.0.133"
 serde_repr = "0.1.20"
 serde_with = "3.11.0"
-sha2 = "0.10.8"
+sha2 = { workspace = true }
 thiserror = { workspace = true }
+num-bigint = "0.4"
```
joe-p, line 25: Think this was added by mistake?

.gitattributes
```diff
@@ -0,0 +1,3 @@
+# Treat text as lf
```
joe-p, line 1: Why do we need to add this now (and haven't before?)

crates/algokit_transact/Cargo.toml
aorumbayev: @PatrickDinh i removed it from toml i assume you added there initially when you started on assumption that abi will be inside transact crate and it was left there after decision was made for a separate crate for abi

.gitattributes
aorumbayev: i assume most of us are developing on linux and macs, this is usually relevant to eliminate inconsistencies with git on windows, @PatrickDinh have you switched to windows recently ? :)

crates/algokit_transact/Cargo.toml
PatrickDinh: oops yes, it should be removed

.gitattributes
PatrickDinh: yes it was me on Windows. It helps that I don't get notified that the `test_data.json` file is changed all the time.

crates/algokit_abi/src/method.rs
```diff
@@ -0,0 +1,550 @@
...
+
+/// Represents an argument in an ABI method.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ABIMethodArg {
+    /// The type of the argument.
+    pub r#type: ABIMethodArgType,
+    /// An optional name for the argument.
+    pub name: Option<String>,
+    /// An optional description of the argument.
+    pub desc: Option<String>,
```
neilcampbell, line 171: What is desc used for? Do we need it?
We've been using the pattern of full names, so potentially this should be `description` if we keep it.

crates/algokit_abi/src/method.rs
```diff
@@ -0,0 +1,550 @@
...
+    pub returns: Option<ABIType>,
+    /// An optional description of the method.
+    pub desc: Option<String>,
+}
+
+/// Represents an argument in an ABI method.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ABIMethodArg {
+    /// The type of the argument.
+    pub r#type: ABIMethodArgType,
```
neilcampbell, line 167: Is it better to use `argType`?

crates/algokit_abi/src/types/collections/array_dynamic.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    pub(crate) fn decode_dynamic_array(&self, value: &[u8]) -> Result<ABIValue, ABIError> {
+        if value.len() < LENGTH_ENCODE_BYTE_SIZE {
+            return Err(ABIError::DecodingError(
+                "Byte array is too short to be decoded as dynamic array".to_string(),
+            ));
+        }
+
+        // The first 2 bytes in the value determines how many values in the array
+        let values_count = u16::from_be_bytes([value[0], value[1]]);
```
neilcampbell, line 45:
```suggestion
        let values_count = u16::from_be_bytes(value[..2]);
```

crates/algokit_abi/src/method.rs
```diff
@@ -0,0 +1,550 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ABIMethod {
+    /// The name of the method.
+    pub name: String,
+    /// A list of the method's arguments.
+    pub args: Vec<ABIMethodArg>,
+    /// The return type of the method, or `None` if the method does not return a value.
+    pub returns: Option<ABIType>,
+    /// An optional description of the method.
+    pub desc: Option<String>,
```
neilcampbell, line 160: What is desc used for? Do we need it?
We've been using the pattern of full names, so potentially this should be `description` if we keep it.

neilcampbell (approved):

crates/algokit_abi/src/method.rs
```diff
@@ -0,0 +1,550 @@
...
+/// Constant for void return type in method signatures.
+const VOID_RETURN_TYPE: &str = "void";
+
+/// Represents a transaction type that can be used as an ABI method argument.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ABITransactionType {
+    /// Any transaction type
+    Txn,
+    /// Payment transaction (algo transfer)
+    Pay,
```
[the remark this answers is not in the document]
aorumbayev: aligned names

PR #218, opened 2025-07-29
feat!: introduces multisignature abstractions in utils

CiottiGiorgio:


crates/algokit_utils/src/multisig.rs
```diff
@@ -0,0 +1,47 @@
+use algokit_transact::{Address, MultisigSubsignature};
+
+#[derive(Debug, thiserror::Error)]
+pub enum MultisigError {
+    #[error("Invalid multisig account: {0}")]
+    InvalidMultisigSignature(#[from] algokit_transact::AlgoKitTransactError),
+}
+
+#[derive(Clone)]
+pub struct MultisigAccount {
```
neilcampbell, line 10: I'm not completely clear how we intend this to be used.
The `MultisigAccount` from utils-ts for example requires either accounts with private keys or signer accounts on construction, giving the account the required info to be able to also sign the transaction. https://github.com/algorandfoundation/algokit-utils-ts/blob/main/src/types/account.ts#L20

PR #224, opened 2025-08-07
chore: sync algod oas spec with upstread 4.2.1 go-algorand release

aorumbayev:
# Primary Changes in algod.oas3.json in 4.2.1 go-algorand release:

. Parameter Reordering: The most significant change is that OpenAPI parameters were reordered to follow standard conventions - path parameters now come before query parameters.
. x-go-type Annotations: Added x-go-type: "basics.Address" annotations to address parameters, which is part of go-algorand's move away from their forked oapi-codegen to the upstream version.
. Description Updates: Minor text changes like adding periods to parameter descriptions.

## Proposed tweaks

- Refreshing oas spec with latest 
- Regenerating algod client 
- Updating breaking tests to reflect changes in parameter ordering

References from upstream repo to changes introduced:
https://github.com/algorand/go-algorand/pull/6375
https://github.com/algorand/go-algorand/pull/6306


crates/algod_client/src/models/asset_params.rs
```diff
@@ -30,7 +30,7 @@ pub struct AssetParams {
     pub creator: String,
     /// \[dc\] The number of digits to use after the decimal point when displaying this asset. If 0, the asset is not divisible. If 1, the base unit of the asset is in tenths. If 2, the base unit of the asset is in hundredths, and so on. This value must be between 0 and 19 (inclusive).
     #[serde(rename = "decimals")]
-    pub decimals: u32,
+    pub decimals: u64,
```
neilcampbell, line 33: This is interesting, given it can fit within a u8.
Looks like everything is becoming a u64.

crates/algod_client/src/models/box_model.rs
```diff
@@ -19,11 +19,11 @@ pub struct Box {
     /// The round for which this information is relevant
     #[serde(rename = "round")]
     pub round: u64,
-    /// \[name\] box name, base64 encoded
+    /// The box name, base64 encoded
     #[serde_as(as = "serde_with::base64::Base64")]
     #[serde(rename = "name")]
     pub name: Vec<u8>,
```
neilcampbell, line 25: FYI I have a generator change to properly handle Base64 values in the OAS for endpoints that support msgpack. They are all bytes values and only base64 when using JSON.

It's unrelated to this PR, however figured it was worth a heads up in case you notice issues.

 however a heads up in case you hit issues.

neilcampbell (approved):

PR #234, opened 2025-08-18
feat: migrate from thiserror to snafu

joe-p:
Migrates all the errors from thiserror to snafu. The main difference is that snafu needs enum variants to be a struct so that's what most of this diff is. This was mostly work done by GPT-5, but I don't think we need to go over it with a fine-toothed comb since we have another story for unifying error handling and any problems can be caught then. For now though, every test is passing. 

This PR won't be merged until the following PRs are merged. Merge conflicts are anticipated and will be resolved in this branch. 

- #218 
- #225 
- #231
- #232
- #233 

Future PRs should use this branch has a base until it is merged


PatrickDinh (approved):
LGTM

PR #237, opened 2025-08-20
docs: added target lib naming convention ADR

lempira:
ADR for the target lib package naming convention

### DIFF
--- a/docs/book/decisions/0001-package-naming-strategy.md
+++ b/docs/book/decisions/0001-package-naming-strategy.md
@@ -1,0 +1,110 @@
+---
+status: proposed
+date: 2025-08-20
+decision-makers: Bruno Martins, Larkin Young, Joe Polny, David Rojas
+consulted: [TBD]
+---
+
+# Package Naming Strategy for Multi-Language AlgoKit Libraries
+
+## Context and Problem Statement
+
+The AlgoKit ecosystem currently includes utilities libraries named `algokit-utils-ts` and `algokit-utils-py` for TypeScript and Python respectively. As we expand to support additional languages (Go, Kotlin, Swift, and Rust) and evolve these libraries into more comprehensive SDKs, we need to establish a consistent naming strategy.
+
+The current "utils" naming was adopted due to naming conflicts with `algosdk` packages published by Algorand Technologies. However, these libraries have evolved beyond simple utilities and now provide comprehensive SDK functionality for Algorand development.
+
+## Decision Drivers
+
+- Clear and consistent naming across all target languages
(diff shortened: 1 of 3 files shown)


docs/book/decisions/0001-package-naming-strategy.md
```diff
@@ -0,0 +1,110 @@
...
+- `algokit-kt` (Kotlin)
+- `algokit-swift` (Swift)
+- `algokit-rs` (Rust)
+
+**Pros:**
+- Clean, simple naming
+- Strong AlgoKit brand association
+- Language designation still clear
+
+**Cons:**
```
PatrickDinh, line 102: Does this also have the similar cons with `algokit-sdk-{lang}` as well, i.e. this is a breaking change?

docs/book/decisions/0001-package-naming-strategy.md
```diff
@@ -0,0 +1,110 @@
...
+- No breaking changes required
+- Existing users familiar with naming
+- No migration effort needed
+
+**Cons:**
+- Misrepresents the true nature of these comprehensive SDKs
+- "Utils" implies limited utility functions rather than full SDK capabilities
+- Inconsistent with the evolving AlgoKit ecosystem strategy
+
+### 2. Adopt SDK naming: `algokit-sdk-{lang}`
```
PatrickDinh, line 55: I like this name the most for the reasons stated in the pros list. An advantage of moving away from the name `algokit-utils-ts` is that it decouples the 2 packages and make migration easier. Taking `lora` as an example. It has `algokit-utils-ts` and `algokit-subscriber` as dependencies. If we release a new version `algokit-utils-ts` without upgrading `algokit-subscriber`, this could cause package conflicts. However, if we release `algokit-sdk-ts`, the user can point to `algokit-sdk-ts` and they can still use the old `algokit-subscriber` if they need to.

docs/book/decisions/0001-package-naming-strategy.md
```diff
@@ -0,0 +1,110 @@
...
+- `algokit-utils-py` (Python)
+- `algokit-utils-go` (Go)
+- `algokit-utils-kt` (Kotlin)
+- `algokit-utils-swift` (Swift)
+- `algokit-utils-rs` (Rust)
+
+**Pros:**
+- No breaking changes required
+- Existing users familiar with naming
+- No migration effort needed
```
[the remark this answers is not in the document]
neilcampbell: >  A codebase can use both utils in old code paths and the core SDK in new code paths.

We could also do something similar without having another package by adding a new namespace to the existing utils package which contains the core SDK.

PR #239, opened 2025-08-20
feat: add app deployer support

neilcampbell:
Add app deployer support.
This PR also fixes a few API related issue discovered whilst testing the feature.

### DIFF
--- a/crates/algokit_utils/src/applications/app_deployer.rs
+++ b/crates/algokit_utils/src/applications/app_deployer.rs
@@ -544,13 +544,13 @@ impl AppDeployer {
         creator_address: &Address,
         ignore_cache: Option<bool>,
     ) -> Result<AppLookup, AppDeployError> {
-        let address_str = creator_address.to_string();
+        let creator_address_str = creator_address.to_string();
         let ignore_cache = ignore_cache.unwrap_or(false);
 
         if !ignore_cache {
             {
                 let app_lookups = self.app_lookups.lock().unwrap();
-                if let Some(cached_lookup) = app_lookups.get(&address_str) {
+                if let Some(cached_lookup) = app_lookups.get(&creator_address_str) {
                     return Ok(cached_lookup.clone());
                 }
             }
@@ -567,7 +567,7 @@ impl AppDeployer {
 
         // Query indexer for apps created by this address
         let created_apps_response = indexer
-            .lookup_account_created_applications(&address_str, None, Some(true), None, None)
+            .lookup_account_created_applications(&creator_address_str, None, Some(true), None, None)
             .await
             .map_err(|e| AppDeployError::IndexerError { source: e })?;
 
@@ -602,7 +602,7 @@ impl AppDeployer {
                         None,
                         None,
                         None,
-                        Some(&address_str),
+                        Some(&creator_address_str),
                         Some(indexer_client::apis::parameter_enums::AddressRole::Sender),
                         None,
                         None,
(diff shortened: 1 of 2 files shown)


crates/indexer_client/src/models/application_params.rs
```diff
@@ -50,17 +53,8 @@ pub struct ApplicationParams {
...
-            extra_program_pages: None,
-            local_state_schema: None,
-            global_state_schema: None,
-            global_state: None,
-            version: None,
-        }
+    /// Default constructor for ApplicationParams
+    pub fn new() -> ApplicationParams {
+        ApplicationParams::default()
     }
```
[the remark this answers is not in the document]
neilcampbell: Generated code.

crates/algokit_utils/src/applications/app_deployer.rs
```diff
@@ -0,0 +1,1279 @@
...
+        &mut self,
+        creator_address: &Address,
+        ignore_cache: Option<bool>,
+    ) -> Result<AppLookup, AppDeployError> {
+        let address_str = creator_address.to_string();
+        let ignore_cache = ignore_cache.unwrap_or(false);
+
+        if !ignore_cache {
+            {
+                let app_lookups = self.app_lookups.lock().unwrap();
```
PatrickDinh, line 552: should we catch and handle the error instead of `unwrap`?

crates/algokit_utils/src/applications/app_deployer.rs
```diff
@@ -0,0 +1,1279 @@
...
+                    if let Some(app_transaction) = &transaction.application_transaction {
+                        if app_transaction.application_id == 0 {
+                            // App creation transaction
+                            app_creation_transaction = Some(transaction);
+                        } else if latest_app_update_transaction.is_none() {
+                            // Latest app update transaction (first non-creation we encounter due to sorting)
+                            latest_app_update_transaction = Some(transaction);
+                        }
+                    }
+                }
```
PatrickDinh, line 650: nitpick: this feels a bit inefficient because it always traverse the entire array while the app create txn is likely at the end and the latest app update is likely at the beginning. Maybe 2 using filters like utils-ts is an acceptable solution here.

crates/algokit_utils/tests/applications/app_deployer.rs
```diff
@@ -0,0 +1,1227 @@
...
+    });
+    let deployment_1 =
+        get_testing_app_deploy_params(&test_account, &metadata, None, None, None, None).await?;
+
+    let result_1 = app_deployer.deploy(deployment_1).await?;
+
+    context
+        .wait_for_indexer_transaction(&match result_1 {
+            AppDeployResult::Create { result, .. } => result.transaction_ids[0].clone(),
+            _ => panic!("Expected Create result"),
```
PatrickDinh, line 676: I have been replacing the `panic!` with `return Err`, what do you think?

PatrickDinh (approved):

crates/algokit_utils/src/applications/app_deployer.rs
neilcampbell: Lock error handling is one of those scenarios when you want to panic, as there is no easy recovery.
The only time this will return an Err, is when the lock is poisoned.

crates/algokit_utils/tests/applications/app_deployer.rs
neilcampbell: Yeah that works, I'll keep consistent.

crates/algokit_utils/src/applications/app_deployer.rs
neilcampbell: Yeah I thought about doing it the same, however because they are on either ends of the array, the filter has to traverse the entire array anyway, so I figured we may as well do in a single iteration.

PR #243, opened 2025-08-25
feat: utils ffi

joe-p:
Still a lot of work to do but making a draft now for visibility before my PTO. 

It should be noted that the types for the http client protocol methods will be missing return types until this PR is merged and release: https://github.com/mozilla/uniffi-rs/pull/2626. I have confirmed functionality is working as expected, so for now we can just ignore any type errors.


crates/algokit_utils/src/transactions/composer.rs
```diff
@@ -1948,22 +1948,23 @@ impl Composer {
 
         // Group transactions by signer
         let mut transactions = Vec::new();
-        let mut signer_groups: HashMap<*const dyn TransactionSigner, Vec<usize>> = HashMap::new();
+        let mut signer_groups: Vec<(Arc<dyn TransactionSigner>, Vec<usize>)> = Vec::new();
+
         for (group_index, txn_with_signer) in transactions_with_signers.iter().enumerate() {
-            let signer_ptr = Arc::as_ptr(&txn_with_signer.signer);
```
joe-p, line 1953: I want to point out that raw pointers in an async method makes it not thread safe (not Send + Sync) which we need for uniffi.

crates/algokit_http_client/src/lib.rs
```diff
@@ -54,7 +54,7 @@ pub struct HttpResponse {
 pub trait HttpClient: Send + Sync {
     async fn request(
         &self,
-        method: HttpMethod,
+        http_method: HttpMethod,
```
joe-p, line 57: See https://github.com/mozilla/uniffi-rs/issues/2628 for why this had to be renamed

packages/python/algokit_utils/tests/test_utils.py
```diff
@@ -0,0 +1,80 @@
...
+)
+from algosdk.mnemonic import to_private_key
+from nacl.signing import SigningKey
+import base64
+import pytest
+
+MN = "limb estate enhance elegant merry worry spell trophy elegant lab truly step enter destroy split leave beach chalk slight they ignore square tower abandon rough"
+SEED_B64: str = to_private_key(MN)  # type: ignore
+SEED_BYTES = base64.b64decode(SEED_B64)
+KEY = SigningKey(SEED_BYTES[:32])
```
joe-p, line 20: The python SDK is poorly documented for how the to_private_key method actually works, but I think this should work

packages/python/algokit_utils/tests/test_utils.py
```diff
@@ -0,0 +1,80 @@
...
+ADDR = "BEMJKX676TZOOWJYMJPOMGIW4UT6S7UDTUUDCUQNKFC42TY6HVKOIWFOYA"
+
+
+class TestSigner(TransactionSigner):
+    @override
+    def sign_transactions(  # type: ignore
+        self, transactions: list[Transaction], indices: list[int]
+    ) -> list[SignedTransaction]:
+        print("Signing transactions")
```
joe-p, line 29: Needs a proper implementation

joe-p:
Closing this PR in favor of the smaller PRs

PR #250, opened 2025-09-04
feat: ts api jinja generator

aorumbayev:
Initial draft on ts api jinja generator and resulting algod/indexer clients based off algo-fetch from awesomealgo


PatrickDinh:
@aorumbayev I noticed that the `approvalProgram` and `clearStateProgram` when calling `getApplicationById` are `string` while the rust version is `Vec<u8>`. Can you please check it?

PatrickDinh:
hi @aorumbayev, I tested the output from this branch and found some possible improvements

- The type field should be number
```
export type TealValue = {
  /**
   * \[tt\] value type. Value `1` refers to **bytes**, value `2` refers to **uint**
   */
  type: bigint;

  /**
   * \[tb\] bytes value.
   */
  bytes: string;

  /**
   * \[ui\] uint value.
   */
  uint: bigint;
};
```
- In `AssetParams`, fields `unitNameB64`, `nameB64` and `urlB64` should be Uint8Array

PatrickDinh (approved):

PR #251, opened 2025-09-04
feat: rust and ts kmd api client; minor jinja generator tweaks

aorumbayev:
- Adds kmd client generation in rust and ts
- Moves localnet dispenser to leverage native kmd client over algokit cli -> cuts down tests suite run from ~10 to ~4 minutes
- Adds kmd related fields to rust client manager 
- KMD upstream api spec includes request bodies in GET requests, many of the associated schemas are empty objects/structs -> generators have been tweak to account for this and to exclude empty request bodies from GET calls 
- Generic fix in the rust and ts generators to skip including content type header to GET/HEAD/DELETE requests, as well as omitting request params if model is an empty object (found a few of such schemas in kmd spec where spec non idiomatically includes request bodies for GET calls where specified schemas are empty objects)


api/specs/kmd.oas3.json
```diff
@@ -0,0 +1,1907 @@
...
+          },
+          "message": {
+            "type": "string",
+            "x-go-name": "Message"
+          },
+          "signed_transaction": {
+            "pattern": "^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$",
+            "type": "string",
+            "format": "byte",
+            "x-go-name": "SignedTransaction"
```
aorumbayev, line 1005: @joe-p  do you know if this is the same signed txn object used in algod? If yes i can align it by adding out own x algokit vendor extension here so generators embed the model from transact crate/package. Otherwise we can keep as is

PatrickDinh (approved):

PR #253, opened 2025-09-07
chore: composer tweaks whilst building TS composer

neilcampbell:


crates/algokit_utils/src/transactions/composer.rs
```diff
@@ -1313,23 +1338,6 @@ impl Composer {
                             assets_count = assets.len();
                         }
 
-                        // Populate boxes at the transaction level
```
neilcampbell, line 1316: There is a guard above which ensures that boxes can't be populated at the transaction level.

PR #259, opened 2025-09-16
fix: go-algorand 4.3.0 access list behaviour hotfixes

aorumbayev:
Gate transaction-level resource population behind ResourcePopulation::Enabled and stabilize behavior on go-algorand 4.3.0+ (access lists).

Update resource_population tests to enable access-list mode, adjust deterministic expectations (e.g., medium box refs).

NOTE: 4.3.0 on go algorand repo https://github.com/algorand/go-algorand/releases/tag/v4.3.0-stable added new access list behaviour but the release does not seem to include updates to algod open api spec (the new [access](https://github.com/algorand/go-algorand/blob/3e5d694bcadd3aaa4d1e30d208f55eec68c5638d/data/transactions/application.go#L144) field in app call txns is present in code but not in the spec for some reason). This pr mainly aims to accomodate hotfixes to patch failing resource population tests due to new behaviour in algod images.

### DIFF
--- a/crates/algokit_utils/tests/transactions/composer/resource_population.rs
+++ b/crates/algokit_utils/tests/transactions/composer/resource_population.rs
@@ -114,13 +114,12 @@ async fn test_accounts_errors_when_resource_population_disabled(
     })?;
 
     let result = composer.send(None).await;
-
     assert!(
         result.is_err()
             && result
                 .unwrap_err()
                 .to_string()
-                .contains("invalid Account reference")
+                .contains("unavailable Account")
     );
 
     Ok(())
@@ -266,14 +265,6 @@ async fn test_boxes_populated_when_resource_population_enabled(
                 BoxReference {
                     app_id: 0,
                     name: vec![109],
-                },
-                BoxReference {
-                    app_id: 0,
-                    name: vec![],
-                },
-                BoxReference {
-                    app_id: 0,
-                    name: vec![],
                 },
                 BoxReference {
                     app_id: 0,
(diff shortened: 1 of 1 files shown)


crates/algokit_utils/src/transactions/composer.rs
```diff
@@ -1378,6 +1395,7 @@ impl Composer {
     fn populate_group_resources(
         transactions: &mut [Transaction],
         group_resources: SimulateUnnamedResourcesAccessed,
+        _access_lists_enabled: bool,
```
PatrickDinh, line 1398: this param doesn't seem to be used

crates/algokit_utils/tests/transactions/composer/resource_population.rs
```diff
@@ -115,13 +115,24 @@ async fn test_accounts_errors_when_resource_population_disabled(
...
     let result = composer.send(None).await;
 
-    assert!(
-        result.is_err()
-            && result
-                .unwrap_err()
-                .to_string()
-                .contains("invalid Account reference")
-    );
+    // On go-algorand >=4.3.0, access lists may allow success without explicit refs.
```
PatrickDinh, line 118: > may allow success

Does it mean that the bahaviour is undeterministic? i.e. it sometimes throws error

crates/algokit_utils/tests/transactions/composer/resource_population.rs
```diff
@@ -115,13 +115,24 @@ async fn test_accounts_errors_when_resource_population_disabled(
...
+    // When Disabled, we assert no auto refs were added if it succeeds; otherwise accept legacy error.
+    match result {
+        Ok(send_result) => {
```
joe-p, line 121: I'm confused on how it's possible that this transaction group succeeds with resource population disabled. If the resource is not defined, I would expect the AVM to throw an error. The existence of (or lack thereof) the access list shouldn't change the fact that resources still need to be defined

crates/algokit_utils/tests/transactions/composer/resource_population.rs
aorumbayev: removed

PR #260, opened 2025-09-19
Refactor transaction sender to match TS implementation

PatrickDinh:
## Proposed Changes

- Simplify the return types of transaction sender
- Update other areas of the code to fit with this change.
  - Note: the app deployer used to rely on transaction sender, in this PR, I updated it to use the composer directly. I haven't removed the transaction sender from the constructor, it will be done in the work to decoupling algorand client from app client.
- Update test


aorumbayev (approved):

crates/algokit_utils/src/clients/client_manager.rs
```diff
@@ -298,15 +300,13 @@ impl ClientManager {
     #[allow(clippy::too_many_arguments)]
     pub fn get_app_factory(
         &self,
-        algorand: Arc<AlgorandClient>,
+        algorand: Rc<AlgorandClient>,
```
PatrickDinh, line 303: Clippy suggested that we should use `Rc` instead of `Arc` because `AlgorandClient` isn't Send + Sync

crates/algokit_utils/src/applications/app_client/mod.rs
```diff
@@ -48,7 +49,7 @@ type BoxNameFilter = Box<dyn Fn(&BoxName) -> bool>;
 pub struct AppClient {
     app_id: u64,
     app_spec: Arc56Contract,
-    algorand: Arc<AlgorandClient>,
+    algorand: Rc<AlgorandClient>,
```
aorumbayev, line 52: I am curious what caused clippy to start emitting the warning to switch arc to rc because its wasn't emitting this warning prior to this refactor. I am also wondering if while clippy aims to improve performance here - it does not account for the fact that we aim to also export them via uniffi. Based on this section https://mozilla.github.io/uniffi-rs/latest/internals/object_references.html#lifetimes it appears that UniFFI expects the underlying Rust types to keep `Arc<T>` fields so ref-counted lifetimes persist past ffi boundary. I wonder if switching to RC will make it impossible to export it via uniffi or require some additional wrapper around it. I also tested out the branch state before the rc to arc commit and running clippy produced no warnings so curious what was the actual warning in your case?

aorumbayev (changes requested):
Just one main comment on Rc to Arc that i am a bit confused about otherwise looks good !

crates/algokit_utils/src/applications/app_client/mod.rs
PatrickDinh: reverted Rc back to Arc. We will investigate this issue in another PR

PR #261, opened 2025-09-21
feat: ffi_uniffi feature for api clients

joe-p:
Add a ffi_uniffi feature to the generated API clients and makes the necessary changes to the templates


crates/algod_client/src/apis/client.rs
```diff
@@ -295,45 +367,71 @@ impl AlgodClient {
...
+
+        #[cfg(feature = "ffi_uniffi")]
+        {
+            result.map(|v| serde_json::to_string(&v).unwrap_or_default())
+        }
+
+        #[cfg(not(feature = "ffi_uniffi"))]
+        {
+            result
+        }
```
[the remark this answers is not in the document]
joe-p: [2b73edc](https://github.com/algorandfoundation/algokit-core/pull/261/commits/2b73edc5797f806d2a061a11ab8e22fad712ca0a)

api/oas_generator/rust_oas_generator/templates/apis/client.rs.j2
```diff
@@ -111,25 +130,48 @@ impl {{ client_type }}Client {
...
+
+        {% if get_success_response_type(operation) == "UnknownJsonValue" %}
+        #[cfg(feature = "ffi_uniffi")]
+        {
+          result.map(|v| {
+            serde_json::to_string(&v).map_err(|e| Error::Serde {
+                message: e.to_string(),
+            })
+          })?
+        }
```
joe-p, line 162: We can't use serde_json::Value over the FFI, so I think serializing to string for unknown objects is the best option?

PatrickDinh: I think this is probably the best option. Otherwise we would have to workout the response types ourselves and manually maintain them, which can be a decent amount of work.

PatrickDinh (approved):

PR #266, opened 2025-09-26
Support heartbeat and state proof transactions (Rust)

PatrickDinh:
Fixes #

## Proposed Changes

-
-
-


joe-p (approved):
LGTM although I did notice some differences in markdown formatting. Not sure what the best way to align on that is but that is outside the scope of the PR

PR #267, opened 2025-09-28
feat: algod_client python package

joe-p:
Creates an `algod_client` package for Python. The main driver for this is to give the Python utils package proper algod methods and types.

It should be noted that I initially wanted to make `algokit_http_client` its own package as well, but that lead to some problems documented here: https://github.com/algorandfoundation/algokit-core/blob/debug/ext_foreign_pointer_not_set/packages/python/algod_client/tests/test_algod.py#L10

The main problem with not having a separate HttpClient package is that all of the API client packages will need their own HttpClient that will be incompatible with others. For now this is fine because we will use a default implementation as seen in this PR anyways. Even without using the default client a user can still make a custom http client implementation, that will just need to make a separate one for each API client (which might be desirable anyway)


packages/python/algod_client/tests/test_algod.py
```diff
@@ -0,0 +1,13 @@
+from algod_client import AlgodClient
+import pytest
+
+
+@pytest.mark.asyncio
+async def test_algod_client():
+    client = AlgodClient(host="http://localhost:4001", token="a" * 64)
```
joe-p, line 7: In the end we get an intuitive way to instantiate the AlgodClient wihthout the developer needing to touch HttpClient at all

PR #271, opened 2025-10-01
feat: app method call support in ffi composer

joe-p:
Adds support for app method calls to the composer

### DIFF
--- a/crates/algokit_utils_ffi/src/transactions/app_call.rs
+++ b/crates/algokit_utils_ffi/src/transactions/app_call.rs
@@ -193,6 +193,12 @@ impl From<RustABIMethod> for ABIMethod {
 pub enum AppMethodCallArg {
     ABIValue(Arc<ABIValue>),
     AppCreateCall(AppCreateParams),
+    AppUpdateCall(AppUpdateParams),
+    AppDeleteCall(AppDeleteParams),
+    AppCallMethodCall(AppCallMethodCallParams),
+    AppCreateMethodCall(AppCreateMethodCallParams),
+    AppUpdateMethodCall(AppUpdateMethodCallParams),
+    AppDeleteMethodCall(AppDeleteMethodCallParams),
 }
 
 impl From<AppMethodCallArg> for algokit_utils::transactions::app_call::AppMethodCallArg {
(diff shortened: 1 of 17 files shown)


crates/algokit_utils_ffi/src/transactions/app_call.rs
```diff
@@ -0,0 +1,624 @@
...
+                app_create_params,
+            ) => AppMethodCallArg::AppCreateCall(app_create_params.into()),
+            _ => todo!(),
+        }
+    }
+}
+
+create_transaction_params! {
+    #[derive(uniffi::Record)]
+    pub struct AppCallMethodCallParams {
```
lempira, line 231: I don't see `AppCreateMethodCallParams`, `AppUpdateMethodCallParams`, `AppDeleteMethodCallParams`. Are they in scope?

crates/algokit_utils_ffi/src/transactions/composer.rs
```diff
@@ -182,6 +182,36 @@ impl Composer {
                 message: e.to_string(),
             })
     }
+
+    fn add_app_create(&self, params: super::app_call::AppCreateParams) -> Result<(), UtilsError> {
```
lempira, line 186: In comparing the composers, i also see these in the `algokit_utils` crate: `add_app_create_method_call`, `add_app_update_method_call`, `add_app_delete_method_call`, `add_app_update`, `add_app_delete`

crates/algokit_utils_ffi/src/transactions/app_call.rs
```diff
@@ -0,0 +1,624 @@
...
+    FfiTransactionSignerFromRust, RustTransactionSignerFromFfi, TransactionSigner, UtilsError,
+};
+
+use algokit_utils::transactions::app_call::{
+    AppCallParams as RustAppCallParams, AppCreateParams as RustAppCreateParams,
+};
+
+use algokit_abi::ABIMethodArg as RustABIMethodArg;
+use algokit_abi::ABIMethodArgType as RustABIMethodArgType;
+
```
lempira, line 19: Should we also add the `ABIDefaultValue` from the core utils

crates/algokit_utils_ffi/src/transactions/app_call.rs
```diff
@@ -0,0 +1,624 @@
+use crate::abi::{abi_type::ABIType, abi_value::ABIValue};
```
lempira, line 1: I am not seeing `ABIReturn`

joe-p: Since we don't have algod client we don't have the full response modeled. We can add ABI return now though if you want

lempira (approved):

PR #272, opened 2025-10-02
Refactor/composer results

PatrickDinh:
Fixes #

## Proposed Changes

-
-
-

### DIFF
--- a/crates/algokit_utils/tests/applications/app_client/send.rs
+++ b/crates/algokit_utils/tests/applications/app_client/send.rs
@@ -35,10 +35,7 @@ async fn test_create_then_call_app(
         )
         .await?;
 
-    let abi_return = result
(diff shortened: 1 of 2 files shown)


joe-p:
Haven't done a proper review yet, but I will say that `primary_result` seems a bit strange to me. Maybe this? 

```rs
#[derive(Debug, Clone)]
pub struct SendAppMethodCallResult {
    /// The result of the primary (last) transaction
    pub result: TransactionResult,
    /// All transaction results from the composer
    pub group_results: Vec<TransactionResult>,
    /// The group ID (optional)
    pub group: Option<Byte32>,
}
```

I think the main thing for me is that `primary_result` seems more verbose and less discoverable for something that is probably going to be the most commonly desired field

PR #273, opened 2025-10-02
Feat/foreign trait testing

lempira:
## Summary

Implements cross-language FFI testing infrastructure using foreign traits to enable Python-side async test orchestration with Rust business logic. This allows Python tests to control async execution while delegating transaction composition to Rust's FFI Composer.

**Key changes:**
- Add foreign trait pattern for cross-language testing:
  - `ComposerTrait` - transaction composition with all transaction types
  - `ComposerFactory` - factory pattern for creating fresh composer instances
  - `TransactionSigner` - async transaction signing interface
  - `TransactionSignerGetter` - retrieve signers by address
  - `AlgodClientTrait` - algod client operations
- Implement Python wrapper (`PythonComposer`) that delegates to Rust FFI `Composer`
- Add comprehensive FFI test coverage for asset freeze txn only (asset freeze/unfreeze workflows)
- Simplify Composer API by storing dependencies internally (algod_client, signer_getter) to eliminate redundant parameter passing

## Test Plan

- [x] Asset freeze/unfreeze cross-language FFI tests pass
- [x] Python async wrapper correctly delegates to Rust Composer


packages/python/algokit_utils/algokit_utils/ffi_composer.py
```diff
@@ -0,0 +1,129 @@
...
+    def __init__(self, algod_client, signer_getter):
+        """
+        Args:
+            algod_client: The concrete AlgodClient FFI object
+            signer_getter: The TransactionSignerGetter implementation
+        """
+        self.algod_client = algod_client
+        self.signer_getter = signer_getter
+
+    def create_composer(self):
```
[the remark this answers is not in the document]
lempira: done

crates/algokit_utils_ffi/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 pub mod abi;
 pub mod clients;
+pub mod tests;
```
joe-p, line 3: We probably should put this behind a feature flag. We don't need to do it in this PR, but maybe add a TODO?

joe-p (approved):
LGTM. Some things we definitely want to change before prod but shouldn't block this PR. 

Also question: Do we want to keep the `test_utils.py` file now that we have the trait-based testing?

crates/algokit_utils_ffi/src/lib.rs
lempira: done

PR #274, opened 2025-10-06
feat: adds typed block endpoint models; minor fixes in generators; fixes to nodely endpoints

aorumbayev:
Adds comprehensive Block schema and SignedTxnInBlock model, enhancing type safety.

Also:
- Adds manual templates specifically for the complex block endpoint response


crates/algod_client/src/models/signed_txn_in_block.rs
```diff
@@ -0,0 +1,96 @@
...
+    pub sender_rewards: Option<u64>,
+    /// [rr] Receiver rewards.
+    #[serde(rename = "rr", skip_serializing_if = "Option::is_none")]
+    pub receiver_rewards: Option<u64>,
+    /// [rc] Close rewards.
+    #[serde(rename = "rc", skip_serializing_if = "Option::is_none")]
+    pub close_rewards: Option<u64>,
+    /// [dt] State changes from app execution.
+    #[serde(rename = "dt", skip_serializing_if = "Option::is_none")]
+    pub eval_delta: Option<UnknownJsonValue>,
```
joe-p, line 62: Don't we already have EvalDelta model? Or is the model different on this endpoint?

crates/algod_client/src/models/signed_txn_in_block.rs
```diff
@@ -0,0 +1,96 @@
...
+
+use algokit_transact::AlgorandMsgpack;
+
+use crate::models::UnknownJsonValue;
+
+/// SignedTxnInBlock is a SignedTransaction with additional ApplyData and block-specific metadata.
+#[serde_as]
+#[derive(Clone, Default, Debug, PartialEq, Serialize, Deserialize)]
+#[cfg_attr(feature = "ffi_uniffi", derive(uniffi::Record))]
+pub struct SignedTxnInBlock {
```
joe-p, line 28: Can you elaborate on why we need a completely new struct and not just add a model that wraps the existing AlgokitSignedTransaciton? Something similar to `SignedTxnWithAD` in go-algorand:

```go
type SignedTxnWithAD struct {
	_struct struct{} `codec:",omitempty,omitemptyarray"`

	SignedTxn
	ApplyData
}
```

crates/algod_client/src/models/signed_txn_in_block.rs
aorumbayev: Yes, we already have the TEAL value-level EvalDelta model. But the block endpoint’s dt is transactions.ApplyData.EvalDelta, a different, larger struct that includes GlobalDelta/LocalDeltas of StateDelta (which use the TEAL value EvalDelta entries), plus logs and inner transactions. I think its better defer to a raw msgpack bytes for now and add an injected typed model later if we need it.

PR #279, opened 2025-10-09
ci: ci and cd jobs for utils ts distribution via verdacio; replacing bun with npm for root semantic release deps;

aorumbayev:
Tested with my tmp instance of verdaccio on a fork.

IMPORTANT: The following VERDACCIO_URL var and VERDACCIO_TOKEN secret must be set on repo prior to merging (cc @lempira )

### DIFF
--- a/.github/workflows/typescript_ci_cd.yml
+++ b/.github/workflows/typescript_ci_cd.yml
@@ -56,11 +56,11 @@ jobs:
 
       - name: Install dependencies (workspace)
         working-directory: packages/typescript
-        run: npm ci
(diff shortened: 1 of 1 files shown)


joe-p (approved):

aorumbayev:
Hey @lempira any chance you can take a look? We would need the secrets set on the repo before we can merge the PR. Otherwise if you need more time for review and verdaccio setup would it be better to then split this pr into CI and CD parts as CI bit can be merged without waiting for verdaccio setup.

PR #281, opened 2025-10-15
Implement account manager abstraction in Rust

PatrickDinh:
Implementation of the account manager abstraction.
- This doesn't support multisig and logicsig account yet.
- Some abstractions for test fixtures such as the `LocalNetDispenser` and `TestAccount` were replaced by the account manager and `SigningAccount`.


crates/algokit_utils/src/clients/testnet_dispenser_client.rs
```diff
@@ -0,0 +1,259 @@
...
+    ///
+    /// # Parameters
+    /// * `params` - Optional TestNetDispenserApiClientParams parameters.
+    ///
+    /// # Returns
+    /// A new `TestNetDispenserApiClient` instance or an error if the auth token cannot be determined.
+    pub fn new(params: Option<TestNetDispenserApiClientParams>) -> Result<Self, DispenserError> {
+        use std::env;
+
+        let auth_token = if let Some(ref p) = params {
```
aorumbayev: If a user passes Some(TestNetDispenserApiClientParams { auth_token: None, ... }), the environment variable will be ignored

aorumbayev (approved):

PR #327, opened 2026-01-16
fix: optional timestamp genesis

joe-p:
Fixes the timestamp in the genesis response not being optional

Needs #323 to merge first


joe-p:
Closing in favor of #332

PR #331, opened 2026-01-31
chore: disable default feature of http client

joe-p:


lempira (approved):
Just to confirm, this was done so that someone needing a specific implementation of the HTTP client doesn't have to pull in the default implementation, right? You would have to opt in to the default implementation now, correct?

joe-p:
Yes exactly. Mainly useful for WASM where you can't use the default client.

PR #332, opened 2026-01-31
fix: optional genesis timestamp

joe-p:


lempira (approved):
Would be good to start using the upstream algokit-oas-generator like utils-ts and utils-py, but this is a separate effort.

PR #334, opened 2026-02-11
feat: algokit-crypto package

joe-p:
This PR introduces the algokit-crypto crate and packages. 

The rust side mirrors the features we have in TS/Py. Namely async traits that can be used to raw sign and generate algorand-specific signers. 

On the FFI side the API is very simple. `algokit_crypto` has a `ed25519_raw_sign` function for signing abritrary data and `algokit_transact` has a `ed25519_sign_transaction` function for signing Transaction objects. 

There are two main reasons the FFI crate doesn't have the async traits:

* If we want to use traits, we need to use objects. This is probmematic for Kotlin because we cannot integrate with the GC
  * See https://github.com/mozilla/uniffi-rs/issues/2819
* The uniffi-generated code for foreign async traits does not work properly with Swift 6
  * See https://github.com/mozilla/uniffi-rs/issues/2458

I believe both of these issues are addressable in uniffi, but rather than waiting to get those changes upstream in uniffi it seems preferable to just implement simple interfaces over the FFI for now. 

Future PRs will add HD support and then eventually we can add more complex interfaces.


crates/algokit_crypto/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+
+# Enables zeroing of sk in the concrete implementation when it is dropped
+zeroize = { version = "1.8", features = ["derive"] }
+
+# Used random seed generation for ed25519
+getrandom = "0.4.1"
+async-trait = "0.1.89"
+
+[dev-dependencies]
+tokio = { version = "1.49.0", features = ["macros"] }
```
PhearZero: LGTM, my IDE was complaining:

```suggestion
tokio = { version = "1.49.0", features = ["macros", "rt"] }
```

PhearZero (approved):

PR #335, opened 2026-02-21
feat: algo25 module

joe-p:
Adds the algo25 module which is used to go from mneomic to seed (and vice versa).

For now this will live under the crypto package.


packages/android/algokit_algo25/build.gradle.kts
```diff
@@ -0,0 +1,53 @@
...
+ * This is a general purpose Gradle build.
+ * Learn more about Gradle by exploring our Samples at https://docs.gradle.org/8.14.2/samples
+ */
+plugins {
+  alias(libs.plugins.android.library)
+  alias(libs.plugins.kotlin.android)
+}
+
+android {
+  namespace = "com.example.algokit_algo25"
```
michaeltchuang, line 13: @joe-p is this supposed to be an example package name?

packages/android/algokit_algo25/gradle/libs.versions.toml
```diff
@@ -0,0 +1,40 @@
...
+[libraries]
+androidx-core-ktx = { group = "androidx.core", name = "core-ktx", version.ref = "coreKtx" }
+junit = { group = "junit", name = "junit", version.ref = "junit" }
+androidx-junit = { group = "androidx.test.ext", name = "junit", version.ref = "junitVersion" }
+androidx-espresso-core = { group = "androidx.test.espresso", name = "espresso-core", version.ref = "espressoCore" }
+androidx-appcompat = { group = "androidx.appcompat", name = "appcompat", version.ref = "appcompat" }
+material = { group = "com.google.android.material", name = "material", version.ref = "material" }
+
+# Compose dependencies
+androidx-compose-bom = { group = "androidx.compose", name = "compose-bom", version.ref = "compose-bom" }
```
michaeltchuang: @joe-p `compose` is for Android UI libraries...is this generated from Rust FFI? I think maybe we can remove them if it's not needed.  Same with `material` and `appcompat`, `expresso` is UI testing

crates/algokit_crypto/src/algo25/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+
+/// Takes a mnemonic and returns the corresponding master derivation key.
+pub fn mnemonic_to_master_derivation_key(
+    mn: &str,
+) -> Result<[u8; SEED_BYTES_LENGTH], MnemonicError> {
+    seed_from_mnemonic(mn)
+}
+
+/// Takes a master derivation key and returns the corresponding mnemonic.
+pub fn master_derivation_key_to_mnemonic(mdk: &[u8]) -> Result<String, MnemonicError> {
```
PhearZero, line 187: Is it just for consistent naming or is this something to be added later? I remember the sdk has a derivation function similar to kmd, I'm guessing this is a stub

PhearZero (approved):
Looks good overall, checksum seems valid and the explicit check for 25 words was a nice touch!

crates/algokit_crypto/src/algo25/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+        .get(..SEED_BYTES_LENGTH)
+        .ok_or(MnemonicError::InvalidSeedLength {
+            expected: SEED_BYTES_LENGTH,
+            found: sk.len(),
+        })?;
+    mnemonic_from_seed(seed)
+}
+
+/// Takes a mnemonic and returns the corresponding master derivation key.
+pub fn mnemonic_to_master_derivation_key(
```
michaeltchuang, line 180: is `master_derivation_key` only used for hd too?  and not algo25?

i needed a way to get Algo25 public key from mnemonic and thought this was it, but AI was saying it's not the right function, but for BIP39 seed

crates/algokit_crypto/src/algo25/mod.rs
joe-p: These functions are here to align with TS/Py which in turn are aligned with algosdk: https://github.com/algorand/js-algorand-sdk//blob/da8f2e6cd616e9430f8bb8d56bb9d25dffe1b98a/src/mnemonic/mnemonic.ts#L164-L181

I assume they exist because the master key and an ed25519 key are different things technically even though they use the same algorithm

crates/algokit_crypto/src/algo25/mod.rs
joe-p: See this comment: https://github.com/algorandecosystem/algokit-core/pull/335#discussion_r2836877124

I'll be adding HD support in the next PR

packages/android/algokit_algo25/gradle/libs.versions.toml
joe-p: This is probably leftover from when we created that first example app, so it's probably safe to remove

joe-p: Removed in [782b1ae](https://github.com/algorandecosystem/algokit-core/pull/335/commits/782b1ae8acd07346236af1058316d066f273417b). Let me know if there's anything I missed (or shouldn't have removed)

crates/algokit_crypto/src/algo25/mod.rs
michaeltchuang: > See this comment: [#335 (comment)](https://github.com/algorandecosystem/algokit-core/pull/335#discussion_r2836877124)
> 
> I'll be adding HD support in the next PR

oh ok

one question then...i thought this crate was only algo25 based off name (e.g. `crates/algokit_algo25/src/lib.rs`...)...will you rename if it includes HD functions too (e.g. `crates/algokit_account/src/lib.rs`...)?  or will hd be it's own crate (e.g. `crates/algokit_hd/src/lib.rs`...)?

packages/android/algokit_algo25/build.gradle.kts
joe-p: What domain should we use?

michaeltchuang: > What domain should we use?

we are setup on mavencentral for `io.github.algorandecosystem` so I think our `aar` structure might be `io.github.algorandecosystem:algokit-*`

what about using `io.github.algorandecosystem.algokit.algo25`?

crates/algokit_crypto/src/algo25/mod.rs
joe-p: Good question... maybe we rename this package to algokit_recovery and it'll have functions for both HD and algo25 mnemonics?

michaeltchuang: > Good question... maybe we rename this package to algokit_recovery and it'll have functions for both HD and algo25 mnemonics?

i think `algokit_account` might be better since if you add the random function...it'll have create account functionality in a way too.  Or `algokit_wallet` / `algokit_address`?  Solana has Seed Vault, maybe `algokit_seed` could work too?

then `algokit_accountdetail` or `algokit_accountinfo` can have balance, rekey flag, ASAs...or anything beyond account keys/addresses?  what do you think?  or maybe ASA becomes it's own crate...i'm open-minded

crates/algokit_crypto/src/algo25/mod.rs
PhearZero: Ahh that's interesting. I wouldn't have expected that, yea it must be a naming decision. I'm guessing KMD does the same

crates/algokit_crypto/src/algo25/mod.rs
joe-p: I ended up just putting the mnemonic stuff in algokit-crypto for now. I figured once we add more functionality we can maybe break it out into seperate packages, but it's not obvious to me how exactly we would split it up right now

https://github.com/algorandecosystem/algokit-core/pull/335/commits/9a283f220818d1744e88b36beb95cf773a2660ee

PR #337, opened 2026-03-22
ci: trigger maven release for new tags

joe-p:
For now releases will be triggered manually. Any time a new tag is push there will be a maven release

Based on https://github.com/michaeltchuangllc/algokit-walletsdk-kmp/blob/main/.github/workflows/publish-sdk-maven.yml


michaeltchuang:
@joe-p you need some sort of `mavenPublishing` or `maven` node block too for each of the packages in `build.gradle.kts`...i think that's what copilot is also referring to

example
https://github.com/michaeltchuangllc/algokit-walletsdk-kmp/blob/main/wallet-sdk-core/build.gradle.kts#L291

also are we merging algo25 into crypto package or keeping it separate?

michaeltchuang:
@joe-p i think you can use the `gradle.properties` file...here were some of the changes i recommended

https://github.com/algorandecosystem/algokit-core-branch/commit/8df1051cd9f4ef81172a18062e150d5a2c115875#diff-bdad18b99d1ac2932a720baa23b9af58dcfb23e3090ad6c9857ace6396318802

michaeltchuang (approved):
since i tested this already works

PR #342, opened 2026-05-19
feat: composer crate

lempira:
Adds two new workspace crates — algokit_composer (Rust) and algokit_composer_ffi (UniFFI) — implementing a stateless transaction composer for the AlgoKit wallet SDK. The composer takes a list of TxnParams variants (Payment, AssetTransfer, AssetOptIn, OnlineKeyReg, OfflineKeyReg) plus a ComposerParams (suggested params + default validity window) and returns built transactions; when more than one entry is supplied, the results are atomically grouped. A separate sign_transactions helper signs index-paired with 32-byte Ed25519 keys, with secret material wrapped in Zeroizing<> so it clears on drop.

The FFI surface is bytes-only (Vec<Vec<u8>> in and out), keeping the boundary stateless per the workspace's hard constraint — any fluent-builder ergonomics live in the caller's language (e.g. a Kotlin-side TransactionComposer { ... } builder that calls compose(list) at .build() time).

Closes walletsdk: #309, #311, #312, #313, #314, #315.

This is broken up into 9 commits:

1. chore(workspace): add algokit_composer + algokit_composer_ffi members — empty crate scaffolds + workspace registration.
2. feat(composer): add CommonTxnParams, TxnParams enum, and ComposerParams types — public data shapes.
3. feat(composer): implement compose() for Payment — compose() skeleton + Payment branch + build_header / parse_address + ComposerError.
4. feat(composer): add AssetTransfer and AssetOptIn handling — AssetTransfer and AssetOptIn variants (opt-in = canonical 0-amount self-transfer).
5. feat(composer): add OnlineKeyReg and OfflineKeyReg handling — keyreg variants + generic bytes_to_fixed helper + collapsed byte-length error variant.
6. feat(composer): auto-group when txn_params.len() > 1 — Transactions::assign_group when result has >1 entry; canonical group ID verified via TransactionGroupMother.
7. feat(composer): add sign_transactions helper — stateless Ed25519 signing (tokio current-thread runtime, Zeroizing<> for secret material); hoists zeroize to [workspace.dependencies].
8. feat(composer-ffi): expose TxnParams, ComposerParams, compose, and sign over UniFFI — full FFI surface (flattened Records, From/TryFrom, AlgoKitComposerError, #[ffi_func] exports returning bytes); polytest catalog + 9 implemented Python tests; registers the package in tools/build_pkgs.
9. chore(scripts): loop test-all.sh Python build over registered packages


crates/algokit_composer/src/composer.rs
```diff
@@ -0,0 +1,409 @@
...
+        .collect::<Result<Vec<_>, _>>()?;
+
+    if txns.len() > 1 {
+        Ok(txns.as_slice().assign_group()?)
+    } else {
+        Ok(txns)
+    }
+}
+
+fn compose_one(p: TxnParams, composer: &ComposerParams) -> Result<Transaction, ComposerError> {
```
joe-p: nit: maybe this could be called compose_transaction?

crates/algokit_composer/src/signer.rs
```diff
@@ -0,0 +1,114 @@
...
+            message: format!("failed to build tokio runtime: {e}"),
+        })?;
+
+    txns.into_iter()
+        .zip(secret_keys.into_iter().map(Zeroizing::new))
+        .map(|(txn, sk)| sign_one(&rt, txn, sk))
+        .collect()
+}
+
+fn sign_one(
```
joe-p: nit: same as above, maybe sign_transaction? `one` just seems a bit weird to me

crates/algokit_composer_ffi/build.rs
```diff
@@ -0,0 +1,4 @@
+fn main() {
```
joe-p: I think we can get rid of this?

joe-p (approved):
Some small nits but otherwise LGTM

crates/algokit_composer/src/signer.rs
lempira: Updated

crates/algokit_composer_ffi/build.rs
lempira: Yeah, I added it following the pattern of algokit_crypto_ffi crate, but we can get rid of it. We might want to get rid of it in the CryptoCrit as well, in the future PR.

crates/algokit_composer/src/composer.rs
lempira: Updated

